PR #48, opened 2013-03-15
Added set_seed for deterministic random number gernation

cstockton:
Change the placers Faker calls Math.rand to use Faker.random.number. This allows overriding Faker.random.number for libraries who wish to do so. This change compliments the set_seed method for deterministic random number generation.


cstockton:
Closing- Thanks for eventually circling around to this, sorry for my late response as well.

PR #77, opened 2014-04-29
Added bower support and fixed amd definition to be compatible with tools like ember-app-kit

daytonn:


tkoomzaaskz:
FYI. In `npm` you need to publish each version separately and afterwardsyou can't change it (unless you `unpublish` a version and you can't revert that). In `bower` you just publish a link to a given git repository, bower automatically scans all tags, they need to conform to `semver` and this is how it resolves version install. The good part of bower is that it gets updated as long as you publish a tag in git repo, the bad part is that the system doesn't guarantee that a specific version will never change (since git owner can remove and re-publish a tag).

Marak:
Closing unless someone can tell me bower isn't working.

PR #129, opened 2014-10-08
require npm package

greenify:
I assume that you just want people to use `require('faker')`

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -22,7 +22,7 @@
 
 ### usage
 
-      var faker = require('./faker');
+      var faker = require('faker');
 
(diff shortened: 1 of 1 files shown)


tkoomzaaskz:
I agree with that. Most of the people will install *faker* via `npm` and they will execute `require('faker')` instead of `require('./faker')`.

Moreover, I can't think of when I would have to run `require('./faker')` - I would have to install the lib via npm, go into the `node_modules` directory and then it would work. But I don't see that make sense.

@greenify good catch, thanks. By the way, you should update the [docs source](build/src/docs.md) instead of the [generated docs](Readme.md) ;)

PR #130, opened 2014-10-09
Added missing street prefix method

YannBrrd:
Added missing street prefix method (usefull for French addresses at least).


Marak:
Looks good.

PR #150, opened 2014-12-09
Add new jobTitle function to name object.

### LINKED ISSUE #148: Add Job Title/Position Generator
Feature Request: Add a generator that generates job titles/positions. A few examples off the top of my head:

Software Engineer
Financial Consultant
Administrative Assistant
etc.

 It would be a good fit for the 'company' category I think.

JKillian:
I had to make a few concessions when doing this.
Ideally I would have added this function under the `company` module,
but I chose to add it to the `name` module instead,
as the data happened to already exist in the locale files under
`name.title`.

However, I felt that the field names in the locale files were confusing,
so their wrappers in the name module are renamed for clarity.

For example, `faker.name.jobArea()` returns a random element from
locale.name.title.level. In my opinion, 'jobArea' was a much better
description of the data than 'level'.

Finally, fixed a bug in one unit test which didn't restore a function
to its original state after watching it.

fixes #148

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ exports.localeFallback = "en";
 exports.definitions = {};
 
 var _definitions = {
-  "name": ["first_name", "last_name", "prefix", "suffix"],
+  "name": ["first_name", "last_name", "prefix", "suffix", "title"],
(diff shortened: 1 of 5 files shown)


tommed:
This doesn't seem to exist on http://faker.hook.io has it been deployed yet?

PR #152, opened 2014-12-22
Feature/uuid generator

jdrew1303:
A lot of projects would use a uuid as the value in their id fields in models. Currently they would have to use a third party lib or roll their own. This implementation would not be suitable for production but is for testing purposes.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -51,6 +51,16 @@ var random = {
         var key = faker.random.array_element(array);
 
         return field === "key" ? key : object[key];
+    },
+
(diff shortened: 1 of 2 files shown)


jdrew1303:
Hi Marak, I was wondering if there was any update on being able to merge?

Marak:
Merged and published as `v2.1.2`

jdrew1303:
Thanks Marak :smiley: I should have a few more for you in the next few weeks _(credit card details, etc)_.

PR #154, opened 2014-12-26
fix random number bug (with {max: 0})

linkkingjay:


tkoomzaaskz:
@linkkingjay can you give more details on the "bug" you are fixing? Can you give an example of an error? You provided no description of what you've done.

linkkingjay:
This issue was fixed in [this pr](https://github.com/Marak/faker.js/pull/218). However, it did not fix the [origin issue](https://github.com/Marak/faker.js/blob/master/vendor/mersenne.js#L266). If `faker` call `mersenne.rand` with `{max: 0, min -100}`, it will return a random number between 0 and 32768.

Marak:
Looks good @linkkingjay. Thank you for your contribution.

PR #157, opened 2015-01-13
Added test to show that russian locale is broken

radzserg:
I'm not really pro in node testing... maybe I'm wrong, but as far as I understood 
```javascript
// test/name.unit.js
    sinon.stub(faker.name, 'firstName').returns('foo');
    var first_name = faker.name.firstName();

    assert.equal(first_name, 'foo');
```
You stub the result. i.e. in any case you'll get 'foo'. This test doesn't make sense for me. 

So I added new test just to demonstrate that loading of locales doesn't really work.

### DIFF
--- a/test/name.unit.js
+++ b/test/name.unit.js
@@ -6,6 +6,16 @@ if (typeof module !== 'undefined') {
 }
 
 describe("name.js", function () {
+
+    describe("firstNameLocale()", function () {
+        it("returns russian random name", function () {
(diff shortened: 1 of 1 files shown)


kovpack:
My PR with Ukrainian locale https://github.com/Marak/faker.js/pull/225 fixes this problem for Russian also.

Marak:
Thanks for the heads-up @radzserg 

If this isn't already fixed, it should be fixed in `master` now and will be in the upcoming release.

PR #160, opened 2015-01-19
Helpers.shuffle creates undefined entries in array

### LINKED ISSUE #159: faker.definitions.lorem.words gets affected by calling things like faker.lorem.sentence
After calling faker.lorem.sentence a couple of 1000 times, the words array almost only contains undefineds

rubenstolk:
... a valid entry in the array, fixes #159


Marak:
Fixed with `v2.1.4`

PR #165, opened 2015-01-25
Update nl.js

jaytaph:
Typo fix and missed province/state for the Netherlands


Marak:
Nice catch!

Thank you.

PR #166, opened 2015-01-28
Update docs.md

toymachiner62:
Updated `faker.name.findName()` to `faker.name.firstName()`

### DIFF
--- a/build/src/docs.md
+++ b/build/src/docs.md
@@ -13,7 +13,7 @@
 
       <script src = "faker.js" type = "text/javascript"></script>
       <script>
-        var randomName = faker.name.findName(); // Caitlyn Kerluke
+        var randomName = faker.name.firstName(); // Caitlyn Kerluke
         var randomEmail = faker.internet.email(); // Rusty@arne.info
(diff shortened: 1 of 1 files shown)


dmamills:
This change is incorrect, `faker.name.firstName()` returns a given name. Where as `faker.name.findName()` provides a given name and a surname, which the comment indicates.

toymachiner62:
Ahh my bad. What comment indicates the difference?

dmamills:
the line you edited shows a full name
`var randomName = faker.name.findName(); // Caitlyn Kerluke`

toymachiner62:
Hmm doesn't seem very intuitive to me. I think it'd be better suited as `fullName` rather than `findName`

tkoomzaaskz:
This PR considers updating an error in the docs. And @dmamills proved there is no error actually ;)

I agree that `fullName` sounds better than `findName`. But if we wanted to give better names for the generators, we need to consider backward compatibility for all those people who use faker.js. So if you still want to suggest new names, submit a new issue.

Issue #126, opened 2014-09-22
Invalid zip codes for Canada


kynax:
Canada uses postal codes instead of Zip codes.
These should be in the format Letter Number Letter Space Number Letter Number (A1A 1A1).
Currently, US zip codes are generated for Canadian addresses.

See http://en.wikipedia.org/wiki/Postal_codes_in_Canada for more information.

If possible, the postal code should also match the generated city.
The Canada Post website lets you find the postal code for a given address: 
https://www.canadapost.ca/cpo/mc/personal/postalcode/fpc.jsf

Marak:
It looks like #173 should solve this. 

Closing issue for inactivity.

PR #173, opened 2015-02-17
Canadian zip codes (postal codes) now formatted properly. Better support for locale based zip code formats.

portse:
The intent of this was to address issue #126. In doing so, I have allowed for zip code formats to be locale based. Also, a user can now pass in their desired zip code format like:

`faker.address.zipCode("####");`
`faker.address.zipCode("?#? #?#");`

If no format is passed in, faker simply uses the format defined by the locale. 

Also, YT is the official abbreviation for Yukon. Reference: http://en.wikipedia.org/wiki/ISO_3166-2:CA.

### DIFF
--- a/build/build/faker.js
+++ b/build/build/faker.js
@@ -18407,7 +18407,7 @@ en_CA.address = {
     "PE",
     "QC",
     "SK",
-    "YK"
+    "YT"
   ],
   "default_country": [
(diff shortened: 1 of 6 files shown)


Marak:
Looks good! Merged.

Thank you for your contribution @portse

PR #174, opened 2015-02-17
Support gender in names

vectart:
This pull request will fix a few problems related to Russian locale:
- `faker.name.firstName()` returns English name
- `faker.name.lastName()` fails because of misprinted `defintions` https://github.com/Marak/faker.js/blob/master/lib/name.js#L20
- `faker.name.findName()` randomly returns male first and female last names and vice verca

In addition, all of these functions will accept `gender` argument.

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ exports.localeFallback = "en";
 exports.definitions = {};
 
 var _definitions = {
-  "name": ["first_name", "last_name", "prefix", "suffix"],
+  "name": ["first_name", "last_name", "prefix", "suffix", "male_first_name", "female_first_name", "male_middle_name", "female_middle_name", "male_last_name", "female_last_name"],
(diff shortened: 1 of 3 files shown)


vectart:
@Marak I'm not going to bother you, but what do you think of these fixes and improvements?

ibrahimyu:
+1
Sooner or later, we will need to generate names based on gender.

kovpack:
+1 Should be merged

kovpack:
One more thing: using correct `prefix` by it's index is completely wrong way: I suppose, there should be `male_name_prefix` and `female_name_prefix` and fallback to `prefix` for those locales that do not yet support that. Many languages have more than 2 prefixes and I have doubts that their order is and will always be correct (anyway, forcing translators to use some predefined order is wrong approach).

Marak:
@vectart -

Looks OK. I have merged into `master` and will continue to review for upcoming release.

vectart:
@Marak thank you for merging!

PR #182, opened 2015-03-13
Commerce implementation

Varedis:
Implements the commerce functionality from the Ruby implementation. Also updates the API documentation to show examples of functions and there uses.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,6 +29,7 @@ exports.date = require('./lib/date');
 exports.random = require('./lib/random');
 exports.finance = require('./lib/finance');
 exports.hacker = require('./lib/hacker');
+exports.commerce = require('./lib/commerce');
 
 var locales = exports.locales = require('./lib/locales');
 
@@ -40,6 +41,8 @@ exports.localeFallback = "en";
 
 exports.definitions = {};
 
+var _stringDefinitions = ["title", "separator"];
+
 var _definitions = {
   "name": ["first_name", "last_name", "prefix", "suffix"],
   "address": ["city_prefix", "city_suffix", "street_suffix", "county", "country", "state", "state_abbr"],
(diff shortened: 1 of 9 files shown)


tkoomzaaskz:
@Varedis good job! I have synced your branch into my local branch to resolve conflicts. There are two things to be discussed:
* you have provided lots of API call examples to extend documentation. The problem is that it's the same PR as the *commerce implementation* and it should be separate. We could merge commerce quite soon, but it's not the case with the docs - e.g. I think there's no sense to write console.log for each line. Moreover, I think we might need a more automatic solution for the docs, might be a docs generator.
* we would have 2 parallel implementations of `color`: `internet.color` returns a hex (`#FF00FF`), `commerce.color` returns a name (`fuchsia`). I think both are needed, but I'm not sure we should leave it this way. The solutions I see are:
  * leave 2 parallel `color` generators in their namespaces
  * rename both: internet.color -> internet.colorHex, product.color -> product.colorName. But in this case we need to provide backwards-compatibility for internet.color
  * merge those functions into one and provide it in both modules. It would accept a paremeter that states, whether it should be a hex or a name.

@Marak what is your opinion on colors (point 2)?

Marak:
@Varedis -

Going through this now. Looks OK, but the PR is a bit disorganized. I will do my best to merge in the new commerce functionality. 

In the future, I would recommend trying to keep your commit history a bit cleaner when opening pull requests up on other projects. Try keeping a working local copy for changes / commits and when you are ready, clone a fresh version of the project and drop your changes on top of that, so you can enter a few clean commits.

Marak:
Merged. Will test a bit more, but looks pretty good!

index.js
```diff
@@ -48,9 +51,18 @@ var _definitions = {
   "hacker": ["abbreviation", "adjective", "noun", "verb", "ingverb"],
   "phone_number": ["formats"],
   "finance": ["account_type", "transaction_type", "currency"],
-  "internet": ["avatar_uri", "domain_suffix", "free_email", "password"]
+  "internet": ["avatar_uri", "domain_suffix", "free_email", "password"],
+  "commerce": ["color", "department", "product_name", "price"]
 };
 
+_stringDefinitions.forEach(function(d) {
```
Marak: @Varedis - Not sure what the purpose of this statement is. Looks a bit out of place. Removed it and commerce tests started to break.

Can you elaborate on the issue here? I'll try digging into it in a bit...

Marak: Okay. I think I've gotten it sorted out. `separator` and `title` properties should be defined the same as rest of definitions. I have modified the object getter mapping locally to allow for string properties.

Should resolve this.

Marak:
I ran into a few issues getting commerce to merge in with the updated codebase.

There are a few coding practices in the pull request I don't prefer.

I've got the commerce code merged into the current version, but it's going to remain commented out for now.

I will try to get it working for the next release. If not, someone else can try to fix it.

Varedis:
@Marak What coding practices in particular?

Marak:
A few little things:

 - `mergeCategories` and `categories` logic should be abstracted somewhere.
 - avoid do / while loops
 - color logic appears to be duplicate / subset of `internet.color`
 - improper handling of default options for min / max values
 - `commerce.price` method appears to be duplicate of `finance.amount` ( both appear to be have bugs )

PR #186, opened 2015-03-17
Fix random number unit test spec

### LINKED ISSUE #185: Invalid random number test spec
for the spec "returns a random number given a maximum value as Object" a value of `10` is valid, but will fail the test. https://github.com/Marak/faker.js/blob/master/test/random.unit.js#L20.

The assert should be checking for `<=` as it does in the other tests.

TheSharpieOne:
Updated the "returns a random number given a maximum value as Object" to include the max value in the check (changed `<` to `<=`)

Previously, this test would randomly fail if the random number ended up being `10`.

Closes #185

### DIFF
--- a/test/random.unit.js
+++ b/test/random.unit.js
@@ -17,7 +17,7 @@ describe("random.js", function () {
 
     it("returns a random number given a maximum value as Object", function() {
       var options = { max: 10 };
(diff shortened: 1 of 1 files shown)


Marak:
Merged.

Thanks!

PR #187, opened 2015-03-18
replace broken shuffle algorithm with one from the internet

jamespharaoh:
The shuffle algorithm was broken, and extremely hard to follow. I've replaced it with one I found on stackoverflow. This kind of thing is notoriously difficult to get right, so best to use a trusted implementation.

To demonstrate the problem, simply call the old shuffle many times. Eventually, all the items in the array are replaced with undefined.

for (var i = 0; i < 10000; i++) { console.log (faker.lorem.words ()); }

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -36,11 +36,31 @@ exports.replaceSymbolWithNumber = function (string, symbol) {
     return str;
 };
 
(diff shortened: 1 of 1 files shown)


Marak:
Fixed with `v2.1.4`

PR #188, opened 2015-03-20
Fixed color copy-paste

jsmirnov:
Tiny fix for internet/color algorithm. Seems to be there was small copy-paste issue.
@Marak


Marak:
Nice catch!

Thank you.

Issue #171, opened 2015-02-10
Add options to the format of internet.userName()


serv:
It would be nice if we allowed passing options to `userName()` that would allow us to generate usernames that just uses `_` or `-` or `.` instead of randomly cycling through all formats.

Marak:
Sounds good!

Will accept pull request for this functionality.

serv:
I'll work on this.

I think creating new methods `userNameUnderscore()`, `userNameDot()` and `userNameHyphen()` are better than updating the current `userName()` method.

It's way easier for people to understand and use the methods, although these methods look repetitive.

Truffula:
I'm not sure if this is too big a departure from the existing API, but for methods that have a number of options (e.g. first name, last name, separator), would it make sense to pass an options object instead of having fixed parameters? e.g: 
```js
fixer.internet.email({
    firstName: 'Bob',
    separator: '.'
});
```

Marak:
I've got a local branch in development that should address this issue as well as #226 #199 and #195

Should have something online in the next day or two.

Marak:
Using `master` branch you should be able to format and combine any existing faker methods.

Examples here: https://github.com/Marak/faker.js/blob/master/examples/generators.js

Sample syntax:

 ```js
console.log(faker.fake('{{name.firstName}}.{{name.lastName}}@{{internet.domainName}}'));
```

I'm thinking this will solve most use cases...let me know.

PR #195, opened 2015-04-13
[feature] internet: added userNameUnderscore/Dot/Hyphen

serv:
Added new internet methods

- `userNameUnderscore`
- `userNameDot `
- `userNameHyphen`

related to #171


FotoVerite:
This seems like a good addition. I am slightly worried about the mound of methods we have going, slightly feeling that having a way to gsub spaces might make more sense then an individual method for each type.

serv:
That's a valid concern. Can you give me an example of how the users would call the 3 cases?

Let me know if you agree with this.

- `userName('.')`
- `userName('-')`
- `userName('_')`

snipesjr:
I feel like adding this would be a good addition, but would it be better to add it like this?


```
 userName: function (firstName, lastName, format) {
        var result;
        firstName = firstName || faker.name.firstName();
        lastName = lastName || faker.name.lastName();

        switch (format ? format : faker.random.number(5)) {
        case 0:
            result = firstName + faker.random.number(99);
            break;
        case 1:
            result = firstName + faker.random.array_element([".", "_"]) + lastName;
            break;
        case 2:
            result = firstName + faker.random.array_element([".", "_"]) + lastName + faker.random.number(99);
            break;
        //userName with dot separator
        case 3:
            result = firstName + "." + lastName;
            break;
        //userName with underscore separator
        case 4:
            result = firstName + "_" + lastName;
            break;
        //userName with hyphen separator
        case 5:
            result = firstName + "-" + lastName;
            break;   
        }
        result = result.replace(/'/g, "");
        result = result.replace(/ /g, "");
        return result;
    }
```

FotoVerite:
serv yes that makes more sense or maybe UserNameFormated.

serv:
Updated

Marak:
Will not merge in current form.

Better to use a DSL and pattern matcher than using formats with arbitrary number values.

For example:

`userName('firstName.lastName')`

or

`userName('firstName-firstName--lastName')`

This seems related to https://github.com/Marak/faker.js/issues/199

Best to figure out a good system for mixins / custom formats that can be applied across the entire library.

Issue #170, opened 2015-02-05
Can't find documentation?


geoguide:
Maybe I'm just being oblivious, but I can not find documentation for faker.js I'm looking for how to use some of the non-obvious methods. For instance a method to give me one of a number of values in an array. Randomize? Or does that just mix it up? Sorry if I'm just missing it, but if it doesn't exist perhaps I could help write it.

Thanks.

dmamills:
To get a random element from an array with faker you can the function `randomize` which is just an alias for `array_element`.

```
var a = [1,2,3,4,5,6,7,8,9,10];
var randA = faker.helpers.randomize(a);
var randB = faker.random.array_element(a);
```

Marak:
@geoguide -

It would be nice to have better documentation for the project.

If you have any suggestions or updates please put them together in a pull request. We will accept improvements to documentation.

Thanks!

tkoomzaaskz:
I'd definitely start with replacing the great one-liner of `<ul><li>...</li></ul>` in Readme.md with markdown list syntax. Adding new generators shoudl include updating the docs.

And +1 for @dmamills - I agree, that the docs could be better.

Anyway, I see this library extremely useful, but the speed of development / accepting PRs is rather slow. @Marak do you consider involving new people to accept contributions for thisproject? I would contribute myself, but I don't want to wait 3 months for a response (and this is what I can see, basing on severalissues I've taken a look at).

dmamills:
Improving the docs would be awesome!

I agree with @tkoomzaaskz. I use this library for almost all of my javascript projects. I'd love to be able to help support this project if you'd be open to letting contributors help keep things moving along a bit faster. I know I'd be completely respectful to any visions or guidelines you have for this library.

tkoomzaaskz:
So it's already two of us :beers:

Marak:
I'm trying my best to keep up with the contributions and make several stable releases per year.

It's hard for me because I haven't been getting any revenues for sometime. I have no money. As much as I'd love to spend a few hours a week to get faster releases out, at the time I literally can't afford to do it.

The best way to keep the project moving is to help answer incoming github issues, review incoming pull requests for problems, and try to alert me a few times if there is a mission critical bug that requires immediate fixing. I'll do my best to keep things moving.

[...]

PR #200, opened 2015-05-08
replacing <ul/li> tags with markdown syntax nested list #170

tkoomzaaskz:
It will be a lot easier to apply and track changes in such list, comparing to the long HTML oneliner. Moreover, I'd replace all HTML will markdown, but it's not that important.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -28,6 +28,10 @@
       var randomEmail = faker.internet.email(); // Kassandra.Haley@erich.biz
       var randomCard = faker.helpers.createCard(); // random contact card containing many properties
 
+### CLI ( Command Line Tool )
+
+see: https://github.com/lestoni/faker-cli
+
 ### Localization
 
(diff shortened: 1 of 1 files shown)


Marak:
@tkoomzaaskz -

Cannot merge this.

The documentation is auto-generated based on the `gulpfile.js`. 

see: 
  https://github.com/Marak/faker.js/blob/master/build/gulpfile.js#L73
  https://raw.githubusercontent.com/Marak/faker.js/master/build/src/docs.md

If you want to make changes to the documentation, you'll need to update the source files, not the generated output.

tkoomzaaskz:
Damn, you're right. Haven't seen the gulpfile there.

By the way, as I was looking over many PRs, it seems that nobody neither updates the docs nor try to rebuild the docs - main readme remains untouched. People can't see it. I would add a contributing.md file where I'd state the rules of contribution. How about that? And I'll move my changes to source docs.

PR #211, opened 2015-05-15
fix: exchange first name and last name in zh_CN.

abruzzihraig:
In Chinese, a last name should be in front of a first name.

### DIFF
--- a/lib/locales/zh_CN.js
+++ b/lib/locales/zh_CN.js
@@ -152,7 +152,7 @@ zh_CN.address = {
   ]
 };
 zh_CN.name = {
-  "first_name": [
+  "last_name": [
     "王",
     "李",
     "张",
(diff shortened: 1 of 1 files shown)


tkoomzaaskz:
@abruzzihraig I'm confused with what you've done ;) you've created a PR and decided it's invalid and dropped it or what? You mentioned a problem (*a last name should be in front of a first name*) and closed it with no comment...

abruzzihraig:
@tkoomzaaskz I'm so sorry for no comment to you. Actually in Chinese, a last name should be in front of a first name, but when I check the code of name.js, I found that there is 'first name' + 'last name' returned. For example a person who has last name '陈' and first name ’思远‘， it should be '陈思远‘ rather than '思远陈', so I closed this pr.
BTW, In Chinese, there should be no space between first and last name, but we can not customize it.

Issue #214, opened 2015-05-27
.words() returns null elements


ndedic:
words() method returns an array which may contain null values. Is this to be expected?
I am able to replicate it using faker.lorem.words() within 100 iterations or less.

{"keywords": faker.lorem.words()}
{"keywords":["molestias","aut",null]}

Thanks,
Nedim

tkoomzaaskz:
Nope, it's not expected :) I checked it and, in fact, undefined values are returned *sometimes*. But the real problem lines in the implementation of the ´Helpers.shuffle´ loop.

Will fix it soon.

ndedic:
Yes, I have managed to get that far but did not have enough time to unwrap .shuffle functionality :-)

tkoomzaaskz:
I'll handle that. Thank you for spotting the bug :)

The fix will be included in the upcoming release.

ndedic:
np

tkoomzaaskz:
@ndedic, the problem is in this line: https://github.com/Marak/faker.js/blob/a6c857f5b442d20182e64cf4f859e114d29a6786/lib/helpers.js#L42
The for loop is hardly readable, but the problem is that the faker.random.number(x) can take [0..x] and if it takes exactly `x` then accessing `x` index of the array will result in returning `undefined` instead of throwing an error (welcome to JavaScript). So if you randomly get the size of the collection in the first step, you'll get exactly one `undefined` in the result.

The code of the `shuffle` - bug-fixed and made more readable - should look like this:

```
function shuffle(o) {
  o = o || ["a", "b", "c"];
  for (var x, j, i = o.length - 1; i > 0; i--) {
    j = faker.random.number(i);
    x = o[i];
    o[i] = o[j];
    o[j] = x;
  }
  return o;
}
```

Anyway, we've just had a quarrel with the owner of this project and as a result all collaborators were removed from this project. Can't understand why, but that's what happened.

sherrman:
Thanks!

PR #215, opened 2015-06-04
Fix issue Marak/faker.js#214 with shuffle()

sherrman:
Fix issue Marak/faker.js#214 with shuffle() sometimes inserting `undefined` into the shuffled array. 

Add a dubious spec that just asserts the proper usage of `faker.random.number()`.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -39,7 +39,7 @@ exports.replaceSymbolWithNumber = function (string, symbol) {
 // takes an array and returns it randomized
 exports.shuffle = function (o) {
     o = o || ["a", "b", "c"];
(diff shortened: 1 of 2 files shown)


Marak:
Nice one, thanks!

nolanamy:
Thanks for fixing this. Was causing occasional test failures for us.

Surprised to still see a one-liner, though – I was a fan of @ducin's suggestion:

```javascript
function shuffle(o) {
  o = o || ["a", "b", "c"];
  for (var x, j, i = o.length - 1; i > 0; i--) {
    j = faker.random.number(i);
    x = o[i];
    o[i] = o[j];
    o[j] = x;
  }
  return o;
}
```
https://github.com/Marak/faker.js/issues/214#issuecomment-105990654

PR #218, opened 2015-06-11
Fixing bug in random.number when max = 0

philberg:
Also added tests for random.array_element

This fixes the same issue noted in [this pr](https://github.com/Marak/faker.js/pull/154)
However, it fixes it a different way.

I came across this issue initially when calling `faker.random.array_element` on an array of length 1.  Since the result of `faker.random.number({max:0})` was not being constrained correctly, the result of the aforementioned command would be `undefined`.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -26,7 +26,7 @@ var random = {
 
         // Make the range inclusive of the max value
         var max = options.max;
-        if (max > 0) {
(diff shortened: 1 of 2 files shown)


Marak:
Looks good, thanks!

pateketrueke:
Yaay! :beers:

PR #224, opened 2015-06-27
Add name interpolation via helpers.template() function

nephridium:
helpers.template() works like mustache(), but checks for #{..} instead
of {{..}}.

Add support for middle_name and fixes "ru" locale which uses male and
female first, middle and last names.

Add unit test for both.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -51,6 +51,14 @@ exports.mustache = function (str, data) {
   return str;
 };
 
+exports.template = function (str, data) {
+  for(var p in data) {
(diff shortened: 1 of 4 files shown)


nephridium:
Hi Marak,
Anything else I need to do?

Marak:
@nephridium -

I've added a new method `Faker.fake` for creating custom faker generators with string formatting.

Example syntax can be found here: https://github.com/Marak/faker.js/blob/master/examples/generators.js

Let me know if this new API solves your needs. I don't think I will be able to merge your pull request in it's current format. 

PR #225, opened 2015-06-27
Add basic Ukrainian support and fix some bugs

kovpack:
Summary:
- add basic **Ukrainian support**;
- fix bug with `faker.name.lastName()` method that was broken for languages like Ukrainian, Russian and maybe some others (added missing definitions, fixed typos in methods etc.);
- localization files are stored in separate `/locales/uk/` directory in separate files (I suppose, this approach is better than one 5000+ line file that is difficult to work with).

Testing and building:
- all tests pass, `gulp` run locally, but built files are not added to PR (I suppose, @Marak should add more extended explanation what should be added, and what should not, because running `gulp` was required in `Readme.md`, but `gulp` creates new build files and adds new methods to the list in `Readme.md` file). So, I suppose, author will run `gulp` locally when bumping new version.


kovpack:
@Marak, are you interested in this PR and support of new locale together with some bug fixes?

Marak:
Yes. Since you put multiple changes into one PR, it will take me longer to go through it.

Will add the new locale.

Thank you.

Marak:
@kovpack - Looks good, merged. Thank you!

In the future, it would be better if you could try to use multiple commits, each being atomic to the functionality that changed.

The preferred way to merge in new functionality for this project is through cherry-picking individual commits.

Regardless of that, thank you for your contribution!

kovpack:
Yeap, clear. Next time will do this way :) I deliberately squashed everything, cause some people prefer one commit instead of multiple.

PR #230, opened 2015-07-06
Username must be a string

xomaczar:
Eliminate cases when result is a number, in which case, replace does not exist.


Marak:
Thanks!

PR #234, opened 2015-07-09
Canadian zip format broken after locale spliting

portse:
This should resolve and ensure the integrity of #173.

Reverted back to simply using the 'postcode' definition, which exists for every locale.


Marak:
Thank you.

Would it be possible to add test coverage to this PR? 

It would be good to add the coverage.

portse:
@Marak Let me know if you think this is sufficient coverage.

Marak:
Looks better. Thank you!

I think overall in the project we'll have to improve general fallback logic for locales.

Thank you for your contributions!

Issue #239, opened 2015-07-13
FR: Allow specification of a seed value


VanTanev:
It would be nice if we could get deterministic output from Faker. It seems that the lib used for random generation already supports this (https://github.com/Marak/faker.js/blob/master/vendor/mersenne.js#L273), so it's just a matter of exposing this interface somewhere on Faker.

Marak:
Sounds good. Will accept a PR.

Marak:
Should be fixed in `master`.

Thanks @Aweary !

PR #243, opened 2015-07-15
Implement faker.seed for deterministic output

Aweary:
This implements https://github.com/Marak/faker.js/issues/239. It uses a prototypical function to re-instantiate the `random` object and passes the seed value all the way to the mersenne function.

To use this method you just call the funciton before generating any of the data.

```js
/* You can use an number */
faker.seed(23085409345);
/* Or you can use an array of numbers */
faker.seed([12, 42, 65, 234, 64]);
var name = faker.name.findName();
```

I added a unit test that covers this in a really basic way. I did verify the files built and passed the tests correctly, but I didn't include those since it would include my username and path from the build process, so you'll want to build it yourself if you merge.


Marak:
@Aweary - Very nice.

Will review and merge this week.

PR #248, opened 2015-07-17
through2 dependency added to devDependencies

kbariotis:
`through2` required by `buid/gulpfile.js` was missing.


Marak:
Thank you!

Issue #250, opened 2015-07-17
New date generators: weekday, month.


mradionov:
```
faker.date.weekday(); // 'Wednesday'
faker.date.weekday({ short: true }); // 'Wed'
faker.date.month(); // 'September'
faker.date.month({ short: true }); // 'Sep'
```
I've recently had been faking some data and felt a need in these "date words", to get them one usually ends up including a library or manually creating arrays of all the weekdays and month names. I wonder if it's gonna be a good addition to a library and won't overload it.

Marak:
@mradionov -

I think it's a pretty good idea!

I would probably create some new data definitions ( in the specific locale ) for date words. Create a few arrays of values such as month and day names. From there you can add a few methods to `date.js` file.

Will accept a pull request for this functionality.

mradionov:
Thanks, @Marak, your suggestions are clear to me and I'd love to make a pull request. One more question: is it required for a contributor to provide translations for all locales when adding a new generator or at least default 'en' would be enough (+ maybe a few more which contributor is aware of) ?

Marak:
Just contribute whichever languages you can. 

`en` is a good one to start with, as others can translate from there.

mradionov:
This feature was released in 3.0.1 and I guess the issue can be closed. Thanks.

PR #251, opened 2015-07-18
Adding generators: faker.date.month(), faker.date.weekday()

mradionov:
According to the issue https://github.com/Marak/faker.js/issues/250 adding date generators:
```
faker.date.month(); // 'September'
faker.date.month({ abbr: true }); // 'Sep'
faker.date.month({ context: true }); // 'September', may be different for other language
faker.date.month({ abbr: true, context: true }); // 'Sep', may be different for other language

faker.date.weekday(); // 'Friday'
faker.date.weekday({ abbr: true }); // 'Fri'
faker.date.weekday({ context: true }); // 'Friday', may be different for other language
faker.date.weekday({ abbr: true, context: true }); // 'Fri', may be different for other language
```

I've used [sources of CLDR](http://unicode.org/cldr/trac/browser/tags/release-27/common/main/en.xml#L1799) as a reference for `en` and `ru` translations and abbreviations.
In russian language (I am sure in some other languages too) words and their abbreviations may sound different in a context - I've added a `context` option for that.


Marak:
@mradionov -

Looks good!

Issue #246, opened 2015-07-17
Add Greek Locale


kbariotis:
I would be more than happy to start a Greek translation PR.

Thx :)

Marak:
Sounds good!

Will accept a pull request for Greek locale.

Marak:
Merged.

Issue #237, opened 2015-07-10
reduce size of git repo and npm package


Marak:
Now that we are including multiple builds for locales, the overall size of the npm package and github repo has increased.

It would be best if we could figure out a better solution to reduce the time it takes to clone faker / install faker from npm.

retrone:
I just realized that Faker is 74 MB on my system. Has there been more thought about the overall size since last July?

It feels a bit optimistic to me to expect that this won't impact people's build processes. At the least, is it possible to move locales into different repos?

Marak:
Don't clone the entire repo if you don't want the full size with history...

Marak:
```
Cloning into 'faker.js'...
remote: Counting objects: 1319, done.
remote: Compressing objects: 100% (1098/1098), done.
remote: Total 1319 (delta 232), reused 1146 (delta 147), pack-reused 0
Receiving objects: 100% (1319/1319), 2.44 MiB | 0 bytes/s, done.
Resolving deltas: 100% (232/232), done.
Checking connectivity... done.
```

**2.44 Megs...learn to git**

retrone:
Rather than getting frustrated and yelling about my git knowledge, let's go back and forth on this. I like faker, and I want to make it better.

My issue is that faker is 70 MB+ when I `npm install` or have it in my `package.json` (that's something I should have specified, and I apologize for not having done so). This is primarily due to the locales, and is 10X the size of any other package I use. I did tests on a git pull, and the size I noticed earlier is similar for the faker folder (in your vein, can I suggest you learn `du -hc` to verify file sizes after doing a git pull?) - am I wrong in thinking that the git clone is showing a gzipped value or something that is excluding a build process?

A separate and worthwhile debate to have is if size matters at all (I'll leave the double entendres aside) - especially for something used in testing and that only manifests itself on the filesystem (as textual data can often be highly compressed for transmission). My bias is to say that something this large does (it's 6-7X larger than most of my modules), especially if a large subset of the user base may only use a single locale (so it may be low cost to providing a few separate builds by locale).

Maybe that's the real debate we should have? I'm also open to being wrong.

fmalk:
In my case, the size issue affects _heavly_ my AWS Lambda packages. It makes deploy painful slow, to the point where uploading signature times out if not done via S3.
My solution was to keep a build separated from what `npm install` downloads, and purge unneeded `/examples`, and `/build/locale`.

Marak:
@fmalk -

[...]

PR #252, opened 2015-07-18
Greek locale translation

kbariotis:
This is the Greek translation as requested at #246.

(@Marak wow lot of noise in the PR)


Marak:
@kbariotis -

Would it be possible to open a new pull request without the build and example files?

As per #237 we've got to reduce noise / size of the git repo.

The idea is to do one build per release. Thanks!

kbariotis:
@Marak better?

Marak:
@kbariotis -

Looks pretty good!

It looks like you've duplicated some data ( such as `internet` and `lorem` ). Is there a reason for this?

kbariotis:
@Marak I believe that those two are globally accepted as is, so I didn't have to translate them. Also, since there is no fallback to another language, it's better to have them than not have them. What you think?

Marak:
@kbariotis - I merged extra data for now. Until I can make a solution upstream for better fallbacks, this is a good solution you made. 

bchavez:
I think the Greek locale isn't complete yet... seems like it's missing an entry in: https://github.com/Marak/faker.js/blob/master/lib/locales.js

Issue #169, opened 2015-02-04
generate card numbers in finance module


awinogradov:
You can use [gist](https://gist.github.com/B-Con/4988902) for this. Maybe if i find some time i do this?

Marak:
Sounds goods

FotoVerite:
Sounds good. I'm curious why you feel it fits best in the finance module?

verybigman:
Because card is on of many payments methods and contain human finance information. @FotoVerite do you think otherwise?

Marak:
@verybigman - Any updates on a pull request for this?

verybigman:
I return to this issue in the next week, man!:)
Чт, 9 апр. 2015 г. в 21:20, Marak <notifications@github.com>:

Marak:
@verybigman - Any updates on a pull request for this?

:-)

verybigman:
@Marak Sorry! I do not have time for it now :(

thebergamo:
Maybe I can do this =D

Marak:
Should be somewhat easy @thebergamo, let us know with a PR if you can.

zpiman:
Ehm... Hello, is this this still alive? I sort of implemented this feature....

awinogradov:
nobody here :( it's already unneeded feature :(

zpiman:
Well I needed it :smile:... From what I understand, the last attempt to add it failed because of library licensing problems...

Marak:
#382 has been merged.

Should have this functionality shipped in the next release.

Closing.

armand1m:
Hi, is this on the docs?

armand1m:
Well, no it isn't. I'm going to use this library in a project now and I'll also submit a pull request adding API details to the README.md

PR #258, opened 2015-07-25
feat(creditcard): Add credit card generator

Mardaneus86:
_One of my first PR's, so please be gentle :sweat_smile:_

Added generator for credit cards by the following issuers: Visa, Mastercard, American Express and Discover for issue #169.

This isn't a perfect implementation, so there is certainly room for improvement. There is still some code duplication but with the added set of unit tests, this can be refactored in an easy way.

For lengths and first digits, I used http://www.freeformatter.com/credit-card-number-generator-validator.html as a reference.
Based on the provided Gist in the issue, I had the first digits of Discover cards fixed to "6011". Other formats are possible in real life (as stated in the code comments).

Sorry to @thebergamo to grab this before you, but I was already working on it for a different project. Please add your comments to this one, or let's compare our solutions!

PS: I had some trouble running the unit tests, I'll look into the issue list later if this has already been reported.

**Usage:**

``` javascript
var card = faker.finance.creditcard();
// or
var visa = faker.finance.creditcard("visa");
// or
var mastercard = faker.finance.creditcard("mastercard");
// or
var americanExpress = faker.finance.creditcard("americanexpress");
// or
var discover = faker.finance.creditcard("discover");
```

This returns an object with the keys "issuer" and "number". Most faker.js API functions return a string in a specific format, but since the issuer and number are so closely related it looked like a good candidate to return an object.


Marak:
Looks good @Mardaneus86 

Will review and merge this week.

Mardaneus86:
Thanks @Marak. Just made some more improvements to the code as I wasn't really happy with how fragile it was. It's now easier to add other card issuers as well.

Looking forward to comments.

Marak:
@Mardaneus86 - 

Looks okay. I'm concerned that there is no license for the gist written by @b-con

I posted a comment at https://gist.github.com/B-Con/4988902

It would probably be best if the fake cc generator was a published module to npm. I think other developers would want that.

Let me know if you can get permission from @b-con or if you can just re-write the module as your npm module.

Marak:
Quick googling found two alternate implementations with licenses:

https://github.com/lourenzo/card-number-generator
https://github.com/VRMink/credit-card-generator

Mardaneus86:
I'll look into those implementations to integrate them into this PR.

Lourenzo's implementation is a fork of VRMink's, they still seem to be quite similar.

Marak:
@Mardaneus86 - Any status on this?

Mardaneus86:
@Marak - Have been quite busy lately, so haven't gotten much progress. However I resumed my work on this yesterday.

I'm currently thinking to use [creditcard-generator](https://www.npmjs.com/package/creditcard-generator) as the library for generating cards (GNU GPL licensed). It supports a few more card types than I've implemented in my initial commit. Card types included: Visa, Mastercard, Amex, Discover, Diners, enRoute, jcb, Voyager. 

My only concern is that this package does not include unit tests. Do you agree with the use of this package even without unit tests?

Mardaneus86:
**Updated usage:**
```javascript
var card = faker.finance.creditcard();
// or
var card = faker.finance.creditcard(cardType);
```

The cardType parameter can be any of the following, case insensitive, strings:
* visa
* mastercard
* amex
* diners
* discover
* enroute
* jcb
* voyager

When you don't provide a card type, a random type will be chosen.
When you provide an unknown card type, it will fall back to Mastercard.

Mardaneus86:
@Marak Any progress on reviewing this PR?

mschinis:
:+1: on merging this.

Marak:
@Mardaneus86 @mschinis 

This looks good, but will require vendor of `credit-card-generator` project.

https://github.com/VRMink/credit-card-generator is listed as GNU GPL, which I think will cause problems with faker license ( MIT ).

Not sure if I can include in library...

mschinis:
best way to find out is to probably contact the owner of that repository, or open a new github issue. I'll just go ahead and do that for you

mschinis:
ah wait. my bad. there's no github issues.

Mardaneus86:
The [original gist](https://gist.github.com/B-Con/4988902) that I based my first implementation on now has a BSD 3-Clause License. If that license doesn't cause issues with the MIT License, I could revert to my initial version? 

That would cause the implementation to go back to only support 4 card types instead of the 8 types supported by [creditcard-generator](https://www.npmjs.com/package/creditcard-generator).

Marak:
Let's try to find the best possible solution here.

My primary goal is getting the best functionality delivered with the minimal amount of code.

I'm not a lawyer, but I'm fairly certain if we statically link or subclass the GPL library it could be considered a derivative work, turning the entire faker project into GPL. I really have no clue. Maybe `require` is considered dynamic linking. I could be wrong about the entire thing.

@VRMink - We'd like to vendor your credit-card generator into faker and subclass it for fake data generation. Would you object to providing a dual license for your work so we can include it? If not, we'll probably end up using a different implementation with an existing MIT license.

Marak:
I've contacted the author of `creditcard-generator` over Github and Email requesting an additional license.

If they are willing to grant it, I will merge the PR as-is.

If not, we'll figure out an alternative solution.

Until we hear back, I will let the PR sit for a bit. If anyone else finds a better code implementation please let us know.

Mardaneus86:
I agree, let's wait and see what the author says.

Hopefully someone can shed some light on the dynamically linking vs derivative work when using it as a npm module. Changing the license for this project because we include something else doesn't sound like the right thing to do.

VRMink:
Hey Guys. Thanks for contacting me. Unfortunately I am not the original author. I got the code from https://github.com/grahamking/darkcoding-credit-card, which has a GPL license, so you need to talk to him, as I just forked the code and change a few bits to better use it with node.js

eliseumds:
Any updates?

Marak:
Merged with #382

PR #263, opened 2015-07-30
Only append suffix to streetName if suffix is present

tcollier:
When `streetSuffix()` returns an empty string, do not append a space to the end of the name part of `streetName()`. With the prior behavior, occasionally, `streetName()` would return something like `"Ashley "` (notice the trailing space), which is undesirable in many circumstances.

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -41,12 +41,17 @@ function Address (faker) {
 
   this.streetName = function () {
       var result;
+      var suffix = faker.address.streetSuffix();
+      if (suffix !== "") {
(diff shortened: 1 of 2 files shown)


Marak:
Should streetSuffix just not be returning an empty string?

tcollier:
There are streets without suffixes, e.g. Broadway in NYC. Having the empty string in that array is a good way to ensure these types of street names will occasionally be returned.

Marak:
@tcollier - Thanks. Looks good.

Merged.

PR #266, opened 2015-08-06
List Greek support to Readme.md

kbariotis:
In addition to #252. Totally forgot it.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -206,6 +206,7 @@ faker.locale = "de";
  * de
  * de_AT
  * de_CH
+ * el_GR
  * en
  * en_AU
  * en_BORK


Marak:
That's odd @kbariotis. The docs should have been updated when the build script was run.

Merged for now. It should fix itself on the next build.

see: https://github.com/Marak/faker.js/blob/master/build/gulpfile.js#L74

bchavez:
@Marak @kbariotis , perhaps because it wasn't fully integrated in:

https://github.com/Marak/faker.js/blob/master/lib/locales.js

and 

https://github.com/Marak/faker.js/tree/master/locale

?

kbariotis:
@bchavez @Marak sorry guys. It's probably my fault. 

Can we make a list of all the files one should change in order to add a locale, and include it in the README? And also let me change all of them in order to integrate the Greek locale.

PR #268, opened 2015-08-13
Remove non-breaking spaces from es_MX locale

### LINKED ISSUE #267: es_MX local is littered with non-breaking spaces
The `es_MX` local is littered with non-breaking spaces `\xA0` in place of regular spaces.  Node.js handles this without issue as it assumes UTF-8 file encoding but for the browser, without explicitly using the `charset` attribute to specify `UTF-8` in the script tag, you will end up with an `Unexpected identifier` error.  A good example is `lib/locales/es_MX/commerce/color.js`.  If I build that local into my sources using Browserify, I get the aforementioned error.  Here is what the inspector shows for the source for `color.js`:

```js
// ...
Â Â  "verde",
Â Â  "azul",
Â Â  "amarillo",
// ...
```

As you can see the `\xA0`, represented here with the `Â`, causes the JavaScript interpreter to freak out.  Of course, I can change my script tag from `<script "myscript.js"></script>` to `<script "myscript.js" charset="UTF-8"></script>` to fix this but should I have to?  Here is a list of all files that are impacted by this, including build artifacts:

```
find . -name "build" -prune -o -name "examples" -prune -o -name "*.js" -exec grep -cH " " '{}' \; | grep -v ":0"
/build/build/faker.js:336
./build/build/locales/es_MX/faker.es_MX.js:336
./examples/browser/js/faker.js:336
./examples/browser/locales/es_MX/faker.es_MX.js:336
./lib/locales/es_MX/commerce/color.js:30
./lib/locales/es_MX/commerce/department.js:21
./lib/locales/es_MX/commerce/product_name.js:53
./lib/locales/es_MX/company/bs_adjective.js:64
./lib/locales/es_MX/company/bs_noun.js:43
./lib/locales/es_MX/company/bs_verb.js:59
./lib/locales/es_MX/team/creature.js:66
```

This is an easy fix but I don't want to do the work if you don't think this is important.

whitlockjc:
Fixes #267


Marak:
@whitlockjc 

Merged.

PR #281, opened 2015-09-14
No space before citySuffix in city name

tullmann:
The citySuffix is a real word suffix (like "ton" or "burgh") so there should not be whitespace before it.

(The cityPrefix is a separate word (like "Lake" or "North"), so still needs whitespace.)


Marak:
@tullmann -

Good catch!

Merged.

PR #282, opened 2015-09-15
Create a method to genarate a string using mask, like jquery mask

norato:
It's useful for creating values like Brazilian CPF and other things

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -130,6 +130,7 @@ This will interpolate the format string with the value of methods `name.lastName
   * contextualCard
   * userCard
   * createTransaction
+  * masked
 * image
   * image
   * avatar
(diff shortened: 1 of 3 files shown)


Marak:
@norato 

This looks good.

Could you quickly elaborate on how this is different from these existing methods:

`helpers.replaceSymbolWithNumber`
https://github.com/Marak/faker.js/blob/master/lib/helpers.js#L36

`helpers.replaceSymbols`
https://github.com/Marak/faker.js/blob/master/lib/helpers.js#L18

I think we might have some overlap? My goal is to provide the best API method for faker, so I'd like to understand exactly which functionality we have, and what we are adding. The existing helper methods seem like they could be refactored / replaced.

chilicoder:
@Marak this will cover the case, I have now
when you need digit OR char on some place

But looks like its just better to add `*` case to `replaceSymbols`.

chilicoder:
I put a PR

Marak:
Closing.

Merge with #524 

I think that will cover the functionality requested in this PR as well.

Issue #283, opened 2015-09-15
faker.internet.domainWord returns domains with apostrophes


rvanider:
While strictly allowed it is very uncommon.

I will supply a pull request with a suggested enhancement.

Marak:
Fixed in master.

PR #284, opened 2015-09-15
Implementation of support for domainWords() without apostrophes

rvanider:
Closes Marak/faker.js#283

This is just one way to implement but it retains the current behaviour unless a user calls faker.internet.allowApostrophe(false);

### DIFF
--- a/lib/internet.js
+++ b/lib/internet.js
@@ -3,6 +3,15 @@ var password_generator = require('../vendor/password-generator.js'),
 
 var Internet = function (faker) {
   var self = this;
+
+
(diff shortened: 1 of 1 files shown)


Marak:
@rvanider @yn5 

Can we get a confirmation of what is and isn't allowed in a web domain?

I'd like to make sure we are actually generating valid domains. We might be able to add more chars, or we might have to remove additional chars.

Marak:
Closing due to in-activity.

This may have been a duplicate of another issue and we removed the word.

If someone encounters this again please let us know.

PR #289, opened 2015-10-06
Added Swedish locale for date

tornqvist:
Added support for Swedish months and weekdays, based on the english implementation.
I also excluded `wide_context` and `abbr_context` as they are optional according to the comments and there is no contextual difference in Swedish.


Marak:
Thanks for your contribution @tornqvist !

PR #291, opened 2015-10-11
hacker.phrase - generate random string for each lexical instance

skylarstein:
hacker.phrase repeats the same string for each lexical type. This commit passes functions instead of strings to faker.helpers.mustache to generate a new random string for each instance. Minor, but hacker.phrase is fun and this reduces repetition in the resulting phrase.

### DIFF
--- a/lib/hacker.js
+++ b/lib/hacker.js
@@ -24,11 +24,11 @@ var Hacker = function (faker) {
   self.phrase = function () {
 
     var data = {
-      abbreviation: self.abbreviation(),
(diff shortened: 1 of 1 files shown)


Marak:
@skylarstein 

Are you aware of anywhere else in the library that suffers from this same issue?

skylarstein:
@Marak 

My initial look didn't uncover any similar issues elsewhere.

Marak:
Merged.

Thank you @skylarstein!

PR #292, opened 2015-10-12
Fix name suffixes

redbugz:
Names were getting prefixes at the end instead of suffixes


akurtz:
I was just about to submit this myself.  Can we get this merged?

Marak:
This should now be cherry-picked into `master` now, not sure why we don't see the merge here..

PR #293, opened 2015-10-13
Lorem.words return string

### LINKED ISSUE #233: "lorem" improvements
Hi,

I was wondering if there is any reason why `lorem.sentence` is not capitalize and does not end with a dot?

> [...] In print or writing, a sentence typically begins with a capital letter and ends with appropriate punctuation.

Also I think it is a bit inconsistent to have `lorem.words` returning an `Array` but `lorem.sentences` and `lorem.paragraphs` returning a `String`.

I notice we can now choose the `separator` for `lorem.paragraphs`. I think the default should not have a `tab`, it is more common.

Regards

pho3nixf1re:
The return value of lorem.words is inconsistent with the other methods as they return strings. It also makes sense that if you are requesting a random value of words it should be a string.

Fixes #233

### DIFF
--- a/lib/lorem.js
+++ b/lib/lorem.js
@@ -15,7 +15,7 @@ var Lorem = function (faker) {
       // strange issue with the node_min_test failing for captialize, please fix and add faker.lorem.back
       //return  faker.lorem.words(wordCount + Helpers.randomNumber(range)).join(' ').capitalize();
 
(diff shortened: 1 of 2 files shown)


Marak:
I hit the same issue actually.

Fixed with https://github.com/Marak/faker.js/commit/5faff98b55a6b3e51862dd3c79c5c0336bcd1953

Regardless, thank you for your contribution. It is much appreciated.

PR #300, opened 2015-11-02
Provide simple way to generate a range of numbers

zpiman:
When a second parameter is added to function `number` make the first parameter `options.min` and the second `options.max`.
This makes it easy to generate a range of values: 

``` javascript
faker.random.number(10,50) // -> generates numbers 10 to 50...
```

### DIFF
--- a/test/random.unit.js
+++ b/test/random.unit.js
@@ -102,6 +102,10 @@ describe("random.js", function () {
   });
 
   describe('arrayElement', function() {
+    it('uses default array if no array is provided', function () {
+      var testArray = ["a", "b", "c"];
+      assert.ok(testArray.indexOf(faker.random.arrayElement()) > -1);
+    });
     it('returns a random element in the array', function() {
       var testArray = ['hello', 'to', 'you', 'my', 'friend'];
       assert.ok(testArray.indexOf(faker.random.arrayElement(testArray)) > -1);


brianmcmillan:
I believe this functionality already exists. From the wiki:
Basic random data

Whole Number
faker.random.number(min,max)
Random number between 0 and "range".
faker.random.number(100); = 92
faker.random.number(5,10); = 9

Decimal number
faker.finance.amount(min,max,decimal places)
Random number between "min" and "max" including decimals to X digits. faker.finance.amount(9000,10000,4); = 9948.8363

zpiman:
Thats what I thought as well @brianmcmillan. But it turns out it is not implemented that way. I was confused in my project why it does not want to work the way it is specified in the wiki :smile:. Then I found that the only way to specify a min value was by passing an option object: `{min:5,max:10}`
I think it is quite a handy feature. That's way I decided to implement it.

Marak:
I think there is a bit of room to improve both of these APIs.

I've noticed we are mutating the incoming `options` in a few places we probably should not.

I'll review this PR a bit more and adjust the existing APIs to make everything work a bit better.

Thank you @zpiman

zpiman:
np :smile:

zpiman:
If you don't mind I can have a look at that @Marak

Marak:
You can look at whatever you please my friend! It's open-source.

zpiman:
Ok let me know what you think

zpiman:
@Marak, have a look at it....

Marak:
Marked for review.

Will have this merged soon.

Thank you for your continued patience @zpiman

lib/random.js
```diff
@@ -10,40 +10,46 @@ function Random (faker, seed) {
...
+        defOptions.max = max;
+        defOptions.min = min_options;
+      } else {
+        // User passed only one number... -> max value
+        defOptions.max = min_options;
+      }
+    } else if (typeof min_options === "object") {
+      // User passed an object -> options to be merged into the defaults
+      for (var i in min_options) {
+        if(defOptions.hasOwnProperty(i) && min_options.hasOwnProperty(i)) {
```
Marak: Why use a `hasOwnProperty` call here instead of referencing the object directly?

Marak:
@zpiman - Can you re-open this commit? It's no showing up as `unknown repository` in Github.

zpiman:
Hmm I might have deleted my fork when I was doing a GitHub clean a while ago... I sort of forgot about this pull request let me see if I still have a copy somewhere...

zpiman:
Ok so I might have deleted the branch on all my local repositories. However this feature would need a better re-implementation anyway... I will look into this tomorrow. I will create a new PR.

grancalavera:
Hi @Marak, I just noticed this `faker.random.number` still mutates its argument. I can fix it if it's still  up for grabs...

PR #301, opened 2015-11-04
add faker.internet.safe_email

alexei-lexx:
Add the `faker.internet.safe_email(firstName, lastName)` method to generate emails with hosts example.com, example.org, example.net.

```javascript
var email = faker.internet.safe_email();
```

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -133,6 +133,10 @@ function Random (faker, seed) {
     return faker.image.image();
   }
 
+  this.locale = function randomLocale () {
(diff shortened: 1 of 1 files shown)


hongkongkiwi:
What about instead of a new method we simply have:
internet.email(options); with adding inside the email method:
options = options || {};
options.safe = options['safe'] || false;

That way we could call internet.email({safe: true}) to make safe emails. This way the email method could be expanded later if we needed to add more options.

Issue #297, opened 2015-10-26
Faker creates invalid phone numbers


TheMcMurder:
faker will occasionally create a phone number like this (faker.phone.phoneNumberFormat(2))  
`1-012-099-8763`

This is completely invalid. Phone Numbers are NXX-NXX-XXXX where N is between 2 and 9 and X is between 0 and 9 see [Wikipedia](https://en.wikipedia.org/wiki/North_American_Numbering_Plan) subsection Numbering system

Marak:
Good catch.

Will accept a pull request to address this.

One thing to consider is Faker supports multiple localities, so there should be separate logic for generating phone numbers from different regions.

greyhwndz:
Is it possible to generate random real city name given a country?

evanjmg:
This is a huge issue. Chance has the same issue

vedmant:
Yeah, it should have custom formats, currently I use just something like this '+4420' + _.shuffle([0,1,2,3,4,5,6,7,8,9]).slice(0, 8).join('')

PR #303, opened 2015-11-11
Edit locales and replaceSymbolwithNumber for more accurate US/CA phone numbers.

shadefinale:
This is a fix for issue #297.

I've written a few tests and made changes to support the `NXX-NXX-XXXX` numbering scheme for US and CA locales by inserting the `!` symbol in the locale to represent a number 2-9.

If there's other locales that need to be changed, or a glaring issue that the tests do not seem to cover (all tests pass), please comment on the PR.


Marak:
@shadefinale 

This is a good fix and I will most likely merge it.

I'd like first like to consider the implications of #333 and if there are other issues which may benefit from a similar style fix to this.

agconti:
Really looking forward to a patch for this issue. Without one, its impossible to use the `phone` provider to test functionality that has validation baked in. 

I see that pr #333 has been closed. Does that allow #303 to be merged in given that #333 is no longer a blocker? 

If there's still more work that needs to be done; I'd be happy help :smile:.

Marak:
Merged.

Thanks!

lib/helpers.js
```diff
@@ -26,6 +26,8 @@ var Helpers = function (faker) {
       for (var i = 0; i < string.length; i++) {
           if (string.charAt(i) == symbol) {
               str += faker.random.number(9);
+          } else if (string.charAt(i) == "!"){
+              str += faker.random.number({min: 2, max: 9});
           } else {
```
bchavez: Hey there. Just wanted to leave a note about this change here as I've been getting [**Bogus**](https://github.com/bchavez/Bogus) C# port caught up to faker.js. Essentially, this change in `replaceSymbolWithNumber` locks characters with `!` to a range of 2-9. I don't think this scales well if we have to introduce a new character inside `replaceSymbolWithNumber ` every time we need a represent a new number range. What if UK (or a different locale) has a different range other than 2-9? A better approach might be to remove `!` from phone number formats and pass phone number formats through [`helpers.regexpStyleStringParse`](https://github.com/Marak/faker.js/blob/master/lib/helpers.js#L144) to define number ranges similar to what's being done in `finance.CreditCardNumber`. For example, instead of phone number formats with `!##-!##-####` they would be `[2-9]##-[2-9]##-####`. That would revert this change here, and let us scale to different locale formats with different ranges without locking down `!` to 2-9.

PR #310, opened 2015-12-01
Support parameters

morrislaptop:
See tests for sample use

### DIFF
--- a/lib/fake.js
+++ b/lib/fake.js
@@ -28,11 +28,19 @@ function Fake (faker) {
 
     // extract method name from between the {{ }} that we found
     // for example: {{name.firstName}}
-    var method = str.substr(start + 2,  end - start - 2);
-    method = method.replace('}}', '');
-    method = method.replace('{{', '');
+    var token = str.substr(start + 2,  end - start - 2);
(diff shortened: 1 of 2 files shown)


morrislaptop:
* bump?

Marak:
@morrislaptop 

This is really nice. I was just starting to implement this myself.

Due to the cross-cutting nature of this change. I will perform an additional review before merging.

Thank you for your continued patience.

lib/fake.js
```diff
@@ -49,7 +57,8 @@ function Fake (faker) {
     var fn = faker[parts[0]][parts[1]];
 
     // replace the found tag with the returned fake value
-    res = str.replace('{{' + method + '}}', fn());
+    eval('var result = fn(' + parameters + ');');
```
Marak: Use of `eval()` is not safe here.

An error in parameter parsing or invalid parameters could cause un-intended consequences to other scopes of the application.

I went ahead and replaced this with a combination of `JSON.parse` and `fn.call`

All tests should be passing.

Marak:
Should be resolved in `master` now.

Thanks @morrislaptop

PR #315, opened 2015-12-09
adjust method to allow for gender based prefixes

dogmatic69:
Mr Jane Smith is ok, but could be better.


Marak:
Good catch @dogmatic69 

This seems correct.

Issue #319, opened 2015-12-24
[feature request] Allow loading additional custom generators


dmelamedcl:
By default `faker` is loading `./lib` which contains all the built-in generators. I would like to add custom generators that I would store in another directory and load them using `faker.load('./custom_lib')`.

Marak:
@dmelamedcl -

If you wanted, you could make a modification to `lib/index.js` where the `Faker` object is being constructed.

https://github.com/Marak/faker.js/blob/master/lib/index.js#L20

You could pass in an additional option as an array of file locations and then iterate on this array and `require` any additional files you may need.

If that isn't good enough, you could try adding a `Faker.prototype.load` method to dynamically require additional modules.

I will probably accept a pull request for either method. 

What I don't want to do is modify the existing require logic for the existing modules ( or generators as you say ).

PR #320, opened 2015-12-26
Add support for loading an external lib

dmelamedcl:
This PR adds the ability to easily load custom modules, once the faker object has been instantiated, using the following syntax:
```
var faker = require('faker')
faker.load('/absolute/path/for/external/modules')
```

Based on feature request https://github.com/Marak/faker.js/issues/319

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -130,4 +130,18 @@ Faker.prototype.seed = function(value) {
   this.seedValue = value;
   this.random = new Random(this, this.seedValue);
 }
+
+Faker.prototype.load = function(_path) {
+  var fs = require('fs');
+  var path = require('path');
+  
+  var items = fs.readdirSync(_path);
+
+  for (var i=0; i<items.length; i++) {
+    var filename = items[i];
+    var module = path.basename(filename, '.js');
+    this[module] = require(_path + '/' + filename)(this);
+  }
+};
+
 module['exports'] = Faker;
--- a/test/all.functional.js
+++ b/test/all.functional.js
@@ -44,3 +44,8 @@ describe("functional tests", function () {
     }
 
 });
+
+describe("load modules", function() {
+  faker.load(__dirname + '/test_lib')
+  assert.ok(faker.test.meaning_of_life() == 42);
+});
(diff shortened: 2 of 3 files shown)


dmelamedcl:
@Marak Can you please take a look?

Marak:
@dmelamedcl -

That looks pretty good.

Would it be possible add some unit tests? Best to keep coverage if we can.

dmelamedcl:
@Marak Added a test

Marak:
Looking good now! Thanks.

@dmelamedcl - Can we assume this will work in node.js environment only?

Do you have any idea how will this behave with a browser build? We might want to remove the `Faker.load` method for the browser build.

dmelamedcl:
@Marak Well, I tried to build the package using the gulp task `browser-package` and it did not complain about anything.

Marak:
What happens if you call `Faker.load` in the browser?

I'm assuming it will fail since it's going to try and require a non-available module?

dmelamedcl:
@Marak I did not try that. I have no problem limiting this feature for node.js. How do you propose to do it?

Marak:
Solution is to remove the method in the build process so it's not included in the browser bundle. Should be a one-liner.

I would like to at least understand what will happen if you call `Faker.load` in the browser though.

dmelamedcl:
@Marak I can try tonight and let you know.

Marak:
Okay. Not a big deal if you can't.

Should be an easy change for the release.

Thanks.

dmelamedcl:
@Marak You were right. Since the function `fs.readFileSync` is not supported by browserify. Can you merge the PR and take care of removing the `load` function in the gulpfile (not exactly sure how to do it)?

dmelamedcl:
@Marak Any news?

dmelamedcl:
?

Marak:
Yes. I will merge this and include it with the next release which should be soon.

Thank you!

Marak:
@dmelamedcl 

I took a second look at this.

What is wrong with just using `require`? I don't see why faker should be responsible for loading custom modules from the file-system.

Issue #318, opened 2015-12-19
[feature request] a non-global Faker instance


john-kurkowski:
It would be nice to be able to instantiate a Faker object with its own state. That way, seeding it doesn't inflict that seed on all other Faker.js users. Maybe they don't want my deterministic output.

Something like this:

```javascript
var myFaker = new window.faker.Faker();
myFaker.seed(123);

var deterministicBsBuzz = myFaker.company.bsBuzz(); // always the same bsBuzz
var stillPseudorandomBsBuzz = window.faker.company.bsBuzz(); // still pseudorandom, unaffected by the seed, above
```

Marak:
@john-kurkowski - 

If you look at the source code, you'll see `Faker` is actually a class which is being initialized for you.

see: https://github.com/Marak/faker.js/blob/master/lib/index.js

If you wanted direct control of creating your own Faker instance you could probably do something like:

https://github.com/Marak/faker.js/blob/master/index.js#L3

Shouldn't require any code changes on our end. Let me know if that works out for you.

john-kurkowski:
Glad the class is already factored out! 

I think there is still a necessary code change: exporting the constructor. #321 is an example of this. Is there another or better way?

Marak:
The `Faker` constructor is already exported.

see: https://github.com/Marak/faker.js/blob/master/lib/index.js#L133

john-kurkowski:
I can see how to get it via a Node.js `require`, yes. But I'm following the README's browser example, trying to plainly execute [build/build/faker.js](https://github.com/Marak/faker.js/blob/v3.0.1/build/build/faker.js). Here are the steps I followed. At the end, I open [this HTML gist](https://gist.github.com/john-kurkowski/7acbad019c9f2b884b82) in my browser. The `console.log` output suggests there's only the lowercase `window.faker` exposed.

```zsh
➜  /tmp  bower i Faker
bower Faker#*                   cached git://github.com/Marak/Faker.js.git#3.0.1
bower Faker#*                 validate 3.0.1 against git://github.com/Marak/Faker.js.git#*
bower faker#~3.0.1             install faker#3.0.1

faker#3.0.1 bower_components/faker
➜  /tmp  wget --quiet https://gist.githubusercontent.com/john-kurkowski/7acbad019c9f2b884b82/raw/5d2d5cdd757b10a3779773ce9cbcb3878a67c1f1/faker-global.html
➜  /tmp  open faker-global.html
```

Marak:
[...]

PR #321, opened 2015-12-27
Export factory method for new Faker instances, for non-global use (#318)

john-kurkowski:
`makeNewInstance()` creates an instance identical to the faker global (same locale data) but mutable state is not shared (e.g. seed value).

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,10 @@
 // since we are requiring the top level of faker, load all locales by default
 var Faker = require('./lib');
-var faker = new Faker({ locales: require('./lib/locales') });
+
+function makeNewInstance() {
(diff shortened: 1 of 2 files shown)


CharlieEarnup:
@john-kurkowski you've got conflicts.

Thanks for writing this up by the way!

john-kurkowski:
🤷‍♂️  no traction on this in years. Last I heard [in #318](https://github.com/Marak/faker.js/issues/318#issuecomment-181133107), @Marak wanted to fix this on his end. Not with this PR. Maybe I should just decline this.

Marak:
Tracking #917

PR #327, opened 2015-12-29
Romanian localization

LoLFactor:
I added the first bits of a Romanian localization.

Don't merge this just yet. I plan to also add dates and, as soon as I pull the data together,  phone numbers and addresses.

I'll post updates over the coming days.


LoLFactor:
I've added Romanian addresses.

LoLFactor:
I've also added Romanian dates and phone numbers. You can merge now.

Marak:
Marked for review.

Will have feedback soon.

Thank you for your patience.

lib/locales/ro/internet/avatar_uri.js
```diff
@@ -0,0 +1,1268 @@
+module["exports"] = [
```
Marak: @LoLFactor -

Is there a reason you have duplicated `internet.avatar_uri` here?

I believe this data is stored in the `en` locale, which all other locales will fallback to for missing methods.

In other words, this method should already be available for the `ro` locale without you having to add it here.

lib/locales/ro/internet/index.js
```diff
@@ -0,0 +1,5 @@
+var internet = {};
```
Marak: This information also appears to be duplicate. 

These methods should already be populated for `ro` locale.

Marak:
@LoLFactor -

Overall commit is OK, but I have issue with locale data for `city`.

The file is really big. I don't think we have any other locales that list such a large amount of cities.

Where did this list come from? Is it every major city in Romania? Could you reduce the list size?

LoLFactor:
@Marak 

I duplicated some data because some modules depend on others (having the same locale). On the test page, some fields would either be blank or in English. That's why I duplicated the data.

As for the cities, yes, that is a list of all the cities in Romania. I guess it can be reduced. 1000 should be enough, yeah?

Marak:
There isn't a test page.

There are the tests, and we have a demo page.

If data isn't populating through the API or the tests, we have to fix the
API.

If the demo page isn't working, that's a separate issue.

Marak:
@LoLFactor -

1000 cities seems OK.

Marak:
Still cannot merge, 16k cities is too much data for the locality.

Marak:
Would like to merge, but is still too much data.

Needs to be reduced in order to merge.

LoLFactor:
Yeah, sorry for that. Totally forgot about this. Will update this weekend.

Marak:
@LoLFactor -

Any updates on this?

I'd like to get this merged, but it's still too much data.

I think 1,000 or so cities is more than enough, maybe even top 100 would be good.

LoLFactor:
Sorry for the horrendous delay. Hope this is OK.

Marak:
Merged.

Very nice, thank you your contributions @LoLFactor.

I would have liked to merge larger amount of data, but we need to try to keep total library size balanced.

PR #330, opened 2015-12-30
Remove Makefile

jory:
As someone coming into the project, the inclusion of a Makefile is really confusing, since it's not actually being used at all.

I added the lint and build tasks to package.json.

Additionally I removed the requirement for istanbul because only the mocha results are being considered.


Marak:
Hrmm. The makefile use to do something.

I'm not using it right now. So let's remove it.

PR #333, opened 2016-01-08
Feature/regex

jdrew1303:
@Marak 

This is a simple evaluation of how it can compress the code down in cases where we want to just generate patterns of information. Another few areas where it would be useful is in phone numbers, zip codes, credit card numbers, etc. It allows for more complex representations than the current string substitution that's in place. Some of the random number generation could also be handled by it.

In the credicard example (Ill upload tomorrow) it removes most of the code. The cvc, card number, dates, all can use the regex to reduce the code down to a few lines.

I haven't heavily refactored the unit tests too much but I do think that by storing the regex separately to the code, similar to the localised data, we can have a lot of the functions generated automatically. It would also allow the library to be moved to another language quite easily in the future.

# Note that this does not have a final build and is not ready to be merged

### DIFF
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -1,6 +1,7 @@
 var Finance = function (faker) {
   var Helpers = faker.helpers,
       self = this;
+      require('randexp').sugar();
 
   self.account = function (length) {
       length = length || 8;
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -3,16 +3,9 @@ var Finance = function (faker) {
       self = this;
 
   self.account = function (length) {
-
       length = length || 8;
-
-      var template = '';
-
-      for (var i = 0; i < length; i++) {
-          template = template + '#';
-      }
-      length = null;
-      return Helpers.replaceSymbolWithNumber(template);
+      var regex = new RegExp("[0-9]{" + length + "}");
+      return regex.gen();
   }
 
   self.accountName = function () {
--- a/lib/internet.js
+++ b/lib/internet.js
@@ -1,5 +1,6 @@
 var password_generator = require('../vendor/password-generator.js'),
     random_ua = require('../vendor/user-agent');
+    require('randexp').sugar();
 
 var Internet = function (faker) {
   var self = this;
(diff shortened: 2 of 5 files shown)


Marak:
@jdrew1303 -

I'm undecided on this.

Your commit is fine and I like the idea of reducing code / making the generated data better.

First off, we will have to vendor `randexp` and `discontinuous-range` libraries. That is OK for me. They seem small.

What I don't like is reading / maintaining Regex. I would like to see instances where using `randexp` library will give us better generated data results. This will also help identify which faker API methods are not returning good data. It looks like you may have already identified some areas which could be improved.

If you could show more concrete improvements or enumerate some specific examples of where faker is generating invalid data, that would help.

Again, I'm undecided on this so far. Mostly because of my dislike of Regex. If we can really improve the core API, then I'm all for it.

jdrew1303:
Hey Marak,

Sorry for the delay in replying. So at the moment I understand what you mean. I picked a lot of the low hanging fruit to change into simple regexs. Ive included a more comprehensive example with credit card demo in PR #341. 

To be honest I dont think that its that its generating invalid data (although in the case of zip codes or phone numbers it may be). The case I'd make for it is that it allows people to concisely express complex cases. Regexs can be fine to maintain if you dont try to do catch all (god) regexs. If you look at the credit card example; I could have used a single regex to express all the combinations, instead I decided to break them up for the sake of making them easier to read and reason about.

Marak:
@jdrew1303 -

I'm going to have to close this for now.

The fact we aren't merging the PR is not a reflection of your work. All of this is good code and good work.

The thing is, it feels like we are sacrificing readability for little gain in functionality. I'm also not sure if we would actually be compressing and saving code since we'd have to add about 400 lines for the deps. `faker` gets built for the browser, so every line counts.

Lastly, I have a personal dislike of Regex, because it's impossible to read and I hate it.

I don't want to seem unreasonable. It looks like some of your regex statements actually improve the integrity of fake date ( such as `ip` or `mac` ) If you'd like to open separate PRs which don't require `randexp` dep, and only modify functions which data will actually be improved, I will merge.

PR #334, opened 2016-01-18
[Locale: en_GB] Added basic support for generating UK postcodes

terrawheat:
Although this doesn't begin to cover the vast array of UK postcodes, it at least provides a vaguely familiar shape of data compared to defaulting to US style ZIP codes. It will occasionally generate outbound codes that don't exist (XY24 9AA would be a possible output for example).

I'm not sure if the Faker DSL for specifying data shape has enough variety to support generating UK postcodes beyond specifying every possibly combination by hand, but I am open to feedback on this subject to make the generation more realistic.

For more info on the craziness of UK postcodes:
https://en.wikipedia.org/wiki/Postcodes_in_the_United_Kingdom


Marak:
Will try to research this a bit more.

Thank you for your contribution @terrawheat

PR #335, opened 2016-01-21
Fix random.uuid not using seeded number generator

john-kurkowski:


Marak:
Good catch @john-kurkowski 

PR #340, opened 2016-02-15
Update Readme.md - added new module that use fakerJs

danibram:
Added mocker-data-generator: https://github.com/danibram/mocker-data-generator


Marak:
ReadMe is auto-generated. 

Updates go into `./build` folder

PR #341, opened 2016-02-16
[WIP] Feature/credit card visa

jdrew1303:
Hey @Marak ,

This is a basic overview of an example where using regex can greatly simplify the code base while providing quite a lot of accuracy in the generated type. There is currently only a single credit card type in there. If you think there would be value in adding this then I can add a lot more types. 

I think this pull request along with #333 and #303 give a good overview of points in the code that can benefit from this approach. I dont think its a silver bullet for all cases, but in this use case it trumps the current model.

Let me know what you think. Also any comments on the API to the module would be welcome.

### DIFF
--- a/lib/credit_card.js
+++ b/lib/credit_card.js
@@ -1,5 +1,8 @@
 function CreditCard (faker) {
-    this.CARDS_SCHEMAS = {
+    var self = this;
+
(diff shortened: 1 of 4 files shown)


jdrew1303:
Based on comments from pull #333 this should also be closed.

PR #348, opened 2016-03-01
Fix 'this' reference in exampleEmail

jmalonzo:
This fixes an issue when using exampleEmail in a fake template (e.g. {{name.exampleEmail}}).


Marak:
Thanks!

PR #350, opened 2016-03-02
[docs] Add jsdoc documentation

ohcibi:
A simple list of all api methods is often not sufficient. Some params
need explanation, some method names are not self explanatory. Plus a
generated documentation can easily be converted into other formats (e.g.
for dash).

For this PR, I've installed jsdoc and added stubs for all existent api
methods. I've tried to annotate all params with their correct type
(there might be mistakes) and wherever there was a comment explaining a
method in the code, I've used that as a description. I also added some
new description to some methods of the address namespace, but I decided
to create a 100% complete API at first when I realized the needed
effort. The default template from jsdoc creates source code listings and
adds links for each method into the source code, which already is pretty
convenient.

I'll continue creating further pull requests, updating the descriptions
when this PR has been accepted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ node_modules/
 # meteor specific
 .build*
 versions.json
+
+doc/
--- a/Readme.md
+++ b/Readme.md
@@ -58,6 +58,9 @@ This will interpolate the format string with the value of methods `name.lastName
 
 ### API Methods
 
+A documentation can be generated with `npm run-script doc`. The docs are put into `./doc`. Here is
+an overview of all api methods:
+
 * address
   * zipCode
   * city
--- a/conf.json
+++ b/conf.json
@@ -1,0 +1,9 @@
+{
+  "opts": {
+    "destination": "./doc/"
+  },
+
+  "plugins": [
+    "plugins/markdown"
+  ]
+}
--- a/lib/address.js
+++ b/lib/address.js
@@ -1,7 +1,18 @@
+/**
+ *
+ * @namespace faker.address
+ */
 function Address (faker) {
   var f = faker.fake,
       Helpers = faker.helpers;
 
+  /**
+   * Generates random zipcode from format. If format is not specified, the
+   * locale's zip format is used.
+   *
+   * @method faker.address.zipCode
+   * @param {String} format
+   */
   this.zipCode = function(format) {
     // if zip format is not specified, use the zip format defined for the locale
     if (typeof format === 'undefined') {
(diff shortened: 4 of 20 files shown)


Marak:
That's nice.

Have you looked at #170 ?

We've been working on an alternate approach.

ohcibi:
Oh I must have overlooked it. Can I dump mine now? 8-). I personally like jsdoc more as I don't actually need a tool to see the information. Plus JetBrains Webstorm and other IDEs have built-in support for it.

Marak:
Who will maintain these JSDOC annotations when the API is updated?

ohcibi:
The one who updates the API? Hows that different with the other approach?

Marak:
So, not you then?

ohcibi:
Well, not necessarily. But again: does nobody has to maintain the docs with the other approach?

Marak:
@ohcibi -

I thought about this a bit, and this is good. Thank you for taking the time to document every method with JSDOC.

The thing is, I don't like the idea of static documentation that lives in-line with the code as comments. I'd much rather have the api contracts live on the `faker` object itself, so it can be enumerating / reflected across other interfaces. We've got most of this in place already and have a bit of tooling using [mschema](https://github.com/mschema/mschema)

As you have pointed out, JSDOC already has many tools available for it.

The best choice for `faker` is to merge this pull request so developers can utilize the existing JSDOC tool-chain. We will also be implementing `mschema` based documentation with #170 

The only down-side of using both JSDOC and mschema will be some additional code clutter, however with our build processes and general structure of the project I feel it won't be too bad. We'll need to make sure the API methods stay updated, but the `faker` API is actually somewhat stable at this point, so it should be managable.

Thanks again for your contribution @ohcibi. I will review and merge this soon.

Marak:
Merged with https://github.com/Marak/faker.js/commit/90a6a04f9cd4a134fec20949e68beb4baf79bfae

Very nice @ohcibi . We should figure out a place to host these html docs now.

re: #170

ohcibi:
@Marak i've setup a github pages site on my fork. Checkout https://ohcibi.github.io/faker.js - so the hosting part is easy. What we need is a proper build chain. I saw `grunt-gh-pages` for example but as you are currently not using any build tool except for npm I didn't want to decide that just by myself.

An approach for an npm-script could be to simply clone the repo with the `gh-pages` branch checked out into the current clone, generate the docs, push back and remove the `gh-pages` clone.

Marak:
@ohcibi -

Looking into this right now. We are actually using `gulp`, so there is a build tool.

Marak:
@ohcibi -

I was able to use the `gulp` toolchain to automate both the generation and publishing of JSDOCs to github pages.

Results: http://marak.github.io/faker.js/

ohcibi:
@marak Great! How will one contribute to this now? Is generating the docs and pushing the branch automated or will it require two prs?

PR #353, opened 2016-03-20
bitcoinAddress fix

ashpr:
* Use faker.random.number
* Fix range. Range was 27 - 36. Should be 27 - 34. Was causing tests to fail.

### DIFF
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -135,6 +135,21 @@ var Finance = function (faker) {
       return symbol;
   }
 
+  /**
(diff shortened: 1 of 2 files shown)


Marak:
Will try again. Currently adjusting the travis settings.

Marak:
@ashpr -

I tried using `finance.bitcoinAddress` and it's generating non-valid bitcoin addresses?

It occurs to me that a "fake" bitcoin address is the same as a real bitcoin address?

PR #360, opened 2016-03-30
Added gendered first names to en locale

landretk:
With the "en" locale selected, using name.firstName() with a gender specified will now return a first name from a list of the top 500 male/female names in the US according to census data. name.findName() also uses this behavior due to it automatically picking a gender if not specified. name.firstName() called without a gender specified falls back to the original behavior for the en locale.

While it is based on US data, I put this change in the en locale because it seems like an expected default feature to me, and the en locale is the default. The en locale contains other US-specific data such as states, so there seemed to be precedent for that.


Marak:
@icebob -

This looks good, but I wonder if we have an issue now with duplicate data and or irregular data coming from the original `firstName` method?

To be clear, how much overlap is there between the legacy `en` names and these new `en` names?

landretk:
I am not sure now my changes to the firstName function from "locale[faker.locale]" to "definitions" on lines 28 and 30 in lib/name.js are correct, but they were needed to pass the tests.

I have no idea how much overlap there is in the data. Maybe the legacy data should be sorted into male and female instead so that it is consistent?

Marak:
@landretk - That might be a good idea for consistency.

Marak:
I looked back into the the project commit history and see a former contributor thought it would be a good idea to merge the `en` first name data ( they use to be separated by gender ). 

There is a list of about 4,000 gender specific names we can switch back to. I might remove our current name data and restore it to the previous commit. I could also merge that with the list you have here.

Marak:
@landretk -

I put together a patch here that had all the names combined, was about 4,800 for female, 1,200 for male. Was taking up something like 60kb of space in the library.

Your commit of top 500 names is much better I think.

Merged. Thank you!

PR #362, opened 2016-04-04
Fix typo in docstring

joshgordon:
Fix typo in docstring for lorem.js. (separatora vs separator)


Marak:
Good catch!

Merged.

Thank you @joshgordon

PR #363, opened 2016-04-04
alphaNumeric takes now count as a argument

krzkaczor:


Marak:
Seems OK.

Merged.

Thank you for your contribution @krzkaczor

PR #366, opened 2016-04-09
Add faker.random.arrayElements that randomly selects a subset of the given array.

joonhocho:
Usage

```
faker.random.arrayElements([1,2,3]) // could be [1], [2], [3], [1, 2], ...

// optional second argument specifies number of elements to return.
faker.random.arrayElements([1,2,3], 2) // could be [1, 2], [1, 3], [2, 3].
```

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -67,6 +67,34 @@ function Random (faker, seed) {
   }
 
   /**
+   * takes an array and returns a subset with random elements of the array
+   *
+   * @method faker.random.arrayElement
+   * @param {array} array
+   * @param {number} count number of elements to pick
+   */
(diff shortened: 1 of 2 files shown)


lib/random.js
```diff
@@ -67,6 +67,34 @@ function Random (faker, seed) {
   }
 
   /**
+   * takes an array and returns a subset with random elements of the array
+   *
+   * @method faker.random.arrayElement
```
Marak: Should be `faker.random.arrayElements`

Marak:
Merged and corrected.

Thanks!

gregwym:
This is really useful. I wonder if there is plan for a new release?

PR #369, opened 2016-04-15
Clean up faker.commerce.department method

koistya:
Remove "max" and "fixedAmount" arguments from the faker.commerce.department method


Marak:
Good catch!

I wonder why these ever got snuck in there. I don't have time to run a blame.

Thank you for your contribution! Merged.

PR #372, opened 2016-04-19
Remove duplicate Congo from country.js

icebob:


Marak:
Thanks @icebob!

Merged.

Issue #339, opened 2016-02-12
Use name formats from definitions in `lib/locales/[locale]/name.js`


Niels-NTG:
Similarly to issue #337 I've noticed that the formatting for names as defined in `lib/locales/[locale]/name.js` get ignored in favour one of two predefined formats in [`lib/name.js`](https://github.com/Marak/faker.js/blob/master/lib/name.js).
```
this.findName = function (firstName, lastName, gender) {
      var r = faker.random.number(8);
      var prefix, suffix;
      // in particular locales first and last names split by gender,
      // thus we keep consistency by passing 0 as male and 1 as female
      if (typeof gender !== 'number') {
        gender = faker.random.number(1);
      }
      firstName = firstName || faker.name.firstName(gender);
      lastName = lastName || faker.name.lastName(gender);
      switch (r) {
      case 0:
          prefix = faker.name.prefix(gender);
          if (prefix) {
              return prefix + " " + firstName + " " + lastName;
          }
      case 1:
          suffix = faker.name.prefix(gender);
          if (suffix) {
              return firstName + " " + lastName + " " + suffix;
          }
      }

      return firstName + " " + lastName;
};
```
This should be replaced by picking a random format from the definitions array of the current locale. [For example](https://github.com/Marak/faker.js/blob/master/lib/locales/nl/name/name.js):
```
module["exports"] = [
  "#{prefix} #{first_name} #{last_name}",
  "#{first_name} #{last_name} #{suffix}",
  "#{first_name} #{last_name}",
  "#{first_name} #{last_name}",
  "#{first_name} #{tussenvoegsel} #{last_name}",
  "#{first_name} #{tussenvoegsel} #{last_name}"
];
```

Marak:
Good catch @Niels-NTG!

I believe a few of the localization fields are being ignored, specifically ones dealing with formatting.

I've now made a note of this and have previously noted #337 

I'll see if I can have both of these issues resolved for the new release.

If you find any other fields or data definitions which are being property utilized, please let me know. We are always looking for ways to improve the faker API.

Thanks for your contributions.

fxnn:
Same for companyName, city, streedAddress, secondaryAddress. Basically you could just grep for locale files using the `#{xyz}` syntax somewhere, because though they are used very often (at least in the german locale), I can't find any mechanism to parse them.

Marak:
@fxnn - You are correct, the parsing code for the additional locale format data was never implemented, just the format themselves was.

[...]

Issue #337, opened 2016-02-05
Improved city name generation


Niels-NTG:
I can't help but notice that the cities Faker generates are almost always constructed of a first name or last name with a typical place name suffix or prefix. I guess this might make sense in most English speaking parts of the world. But very rare in places like The Netherlands, Germany and other nations in continental Europe.

This could be improved by pulling random names from real places from the [GeoNames](http://www.geonames.org/) database. I suppose relying on another service would make Faker less reliable. Alternatively you could have a pre-defined smaller list of names for each language, similar in how the game OpenTTD does its [town names](https://dev.openttdcoop.org/projects/dutchtowns/repository/revisions/2f116ac13084/entry/dtnames.nml) for each language.

Marak:
@Niels-NTG -

I feel like some of the locales actually have city data, maybe not.

If you are able to make a pull request adding arrays of common cities to any of the locales, it would be very much appreciated.

Thank you for your feedback.

Niels-NTG:
I've noticed that faker.js completely ignores the city format defined in `lib/locales/[some locale]/address/city.js` and instead uses a random one defined in this array in [`lib/address.js`](https://github.com/Marak/faker.js/blob/master/lib/address.js):
```
this.city = function (format) {
    var formats = [
      '{{address.cityPrefix}} {{name.firstName}} {{address.citySuffix}}',
      '{{address.cityPrefix}} {{name.firstName}}',
      '{{name.firstName}} {{address.citySuffix}}',
      '{{name.lastName}} {{address.citySuffix}}'
    ];

    if (typeof format !== "number") {
      format = faker.random.number(formats.length - 1);
    }

    return f(formats[format]);
}
```

This is a shame because their already some locales that have different formats defined in their `city.js` files. The France locale for example has [city names defined as `city_name`](https://github.com/Marak/faker.js/blob/master/lib/locales/fr/address/city.js) from the [`city_name.js`](https://github.com/Marak/faker.js/blob/master/lib/locales/fr/address/city_name.js). Due to the current method these entries never get used anywhere.

The format for city names should be pulled from `lib/locales/[some locale]/address/city.js` instead of a predefined array in `lib/address.js`. I've made some attempts making this happen, but haven't been successful thus far. Perhaps somebody with a better understanding of the project structure can do this instead. I will stick with Dutch town names for now.

Marak:
Good catch!

I'll see if I can sort it out.

Marak:
[...]

Issue #232, opened 2015-07-08
Replace hard-coded format arrays used by faker.fake() with definitions


Marak:
Now that we have a proper string generator method `Faker.fake`, we should replace any instances of hard-coded format arrays with a locale backed data definition.

Should be an easy search and replace operation.

PR #373, opened 2016-04-25
Generate by locale definitions

icebob:
Hi @Marak,

when I created the hungarian locale files, I found a problem, that Faker doesn't use format definitions in the locale files. I found many issues about it: #133 #232 #337 #339 

So I made a generate method in the helpers.js, that can solve recursively the locale formats #{...} variables. 

If you need a locale formatted company name use the `faker.helpers.generate("company.name")`. It will use the `company/name.js`, select a random item, and solve the variables. If found e.g. `#{Name.last_name}`, it will use the `name/last_name.js` and select an item...etc while solve every variables.

Examples:

Generate a name by locale formats:

```
faker.helpers.generate("name.name");
```

Generate a street address by locale formats:

```
faker.helpers.generate("address.street_address");
```

Generate a city name by locale formats:

```
faker.helpers.generate("address.city");
```

Generate a company name by locale formats:

```
faker.helpers.generate("company.name");
```

If you think it is useful, merge this PR :)

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -284,7 +284,6 @@ var Helpers = function (faker) {
   self.generateBy = function(def, path) {
     var arr = getProp(def, path.toLowerCase());
     var item;
-    debugger;
     if (typeof(arr) == "string")
       item = arr;
     else if (arr && arr.length > 0)
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -242,6 +242,70 @@ var Helpers = function (faker) {
     };
   };
 
+  function stringToPath(string) {
+    /** Used to match property names within property paths. */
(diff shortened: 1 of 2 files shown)


Marak:
@icebob -

Thank you!

This looks like a great start.

How could we make this work automatically through the existing faker API while calling any method thats scoped to a specific locale?

In order words, how could we make something like `Faker.address.city` method use actual locale formatting data?

see: https://github.com/Marak/faker.js/blob/master/lib/address.js#L45

icebob:
I'm thinking on it. _I made a changes, and now accept index too in path._

Simple method is to call this generate method:
```
this.city = function(format) {
  return faker.helpers.generate("address.city[" + format + "]");
}
```

icebob:
But you can't use this method in name.findName() because there are other parameters what fixing parts of name or gender.

Issue #368, opened 2016-04-13
Faker.fake("{{random.uuid}}") throws error in version 3.1 


donpark:
`Faker.fake("{{random.uuid}}")` throws:

```
TypeError: self.number is not a function
    at replacePlaceholders .../node_modules/faker/lib/random.js:93:29)
```

but not in version 3.0.1.

Calling `Faker.random.uuid()` directly also works.

String format version may be mishandling `this` binding in version 3.1.

skylarstein:
Fixed in PR https://github.com/Marak/faker.js/pull/374

Marak:
Should be resolved with #377 

Published in latest version of faker

PR #374, opened 2016-04-28
[fix] Faker.fake("{{random.uuid}}") throws error in version 3.1

skylarstein:
Fix for https://github.com/Marak/faker.js/issues/368

No other Faker.fake("{{random.X}}") values had this issue.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -90,7 +90,7 @@ function Random (faker, seed) {
       var self = this;
       var RFC4122_TEMPLATE = 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx';
       var replacePlaceholders = function (placeholder) {
-          var random = self.number({ min: 0, max: 15 });
+          var random = faker.random.number({ min: 0, max: 15 });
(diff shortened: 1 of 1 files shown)


john-kurkowski:
The error was introduced in #335. Although `window.faker` looks like a stateless global, it actually has internal state via `faker.seed(...)`. Hence my change to use `this.number`.

Rather than undoing #335 in this PR, can `Faker.fake` thread the `this` context through to `replacePlaceholders`?

francescomalatesta:
Hi guys, having the same issue here. What is the situation right now? Any decisions about this PR? :) thx

ifraixedes:
This should be merge, as it's a bug and the PR fix the issue

Marak:
Marked for fixing in the next release.

Since `Faker.faker` is a sugar syntax for the core methods, you can always work-around it by calling the core methods directly.

orangewise:
@Marak any idea on when the next release is due? Would be nice to have this PR merged :)

ffffranklin:
Proposed fix #402 should handle both usecases.

Marak:
Should be resolved with #377 

Published in latest version of faker

PR #377, opened 2016-05-05
Bindall methods

### LINKED ISSUE #376: Make public API context-agnostic
Current, many of the functions exposed in this lib must be called within the correct context. I'd like to be able to call them in a more free form for composability. Example:

``` js
[1, 2, 3].map(faker.random.uuid);
//=> TypeError: self.number is not a function

// Alternatively: .map(faker.random.uuid.bind(faker.random))
[1, 2, 3].map(() => faker.random.uuid());
//=> [ 'e2600822-2d54-4fe9-9017-af817324f769',
//=>   '5f7eb1a0-02be-4dfb-a1b3-48ae803233b8',
//=>   'ffc0e2a6-69ae-4b69-856f-43b0a09fc6be' ]
```

I'd be happy to submit a PR for this. However, since it's potentially a breaking change (one would be unable to redefine the `this` binding), I just wanted to get a +1 on it before continuing down that path.

jimf:
Statically bind the methods on all modules to their correct context so that the methods can be called in a composable, callback-oriented manner.

I've also updated the functional tests to cover _all_ modules/methods.

Please let me know if I've missed anything (wasn't sure whether the built files should be included in the PR, or just the files I personally touched), or if you'd like anything changed stylistically.

Closes #376

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -38,52 +38,61 @@ function Faker (opts) {
 
   self.definitions = {};
 
+  function bindAll(obj) {
+      Object.keys(obj).forEach(function(meth) {
+          if (typeof obj[meth] === 'function') {
+              obj[meth] = obj[meth].bind(obj);
+          }
+      });
+      return obj;
+  }
+
   var Fake = require('./fake');
   self.fake = new Fake(self).fake;
 
   var Random = require('./random');
-  self.random = new Random(self);
+  self.random = bindAll(new Random(self));
   // self.random = require('./random');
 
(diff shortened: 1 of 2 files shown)


Marak:
This seems OK.

I'm going to want to review this more since it's so cross-cutting.

It might also be a good time to think about how the modules are structured to see if we can make any additional improvements.

Thanks!

Marak:
Merged and published with `v4.1.0`

resolves #402 #368 #335 #374 

Thanks @jimf

PR #378, opened 2016-05-05
Create faker.address.nearbyGPSCoordinate() method

vivekseth:
This method returns a random GPS coordinate that is within a specified distance of the center GPS coordinate. 

It has three parameters: 
coordinate: the center GPS position (if no coordinate given, a random coordinate is returned)
radius: distance from the center position from which to return a GPS coordinate. (default: 10.0)
isMetric: whether to interpret `radius` as miles or kilometers (default: false)


Marak:
That's cool!

I'll have to test a bit to see if it's working.

Can you add any unit tests?

vivekseth:
Added unit tests, and changed the heuristic i'm using. Points returned will now be in the circular region defined by the input coordinate and radius, but they will be concentrated in the center. 

In the future i'm thinking we can allow users to choose between either a normal or uniform distribution for the generated points. Not sure how to do that math at this second, so leaving it for later.

vivekseth:
Anything I can do to help you get this change merged?

Marak:
Will review as part of next release cycle. 

Please see README for more details.

Marak:
@vivekseth -

The `addess.nearbyGPSCoordinate` method has been merged into master.

This is a neat function! Let us know if you have any improvements or additional changes.

Thanks again for your contribution!

PR #379, opened 2016-05-15
Fix misspelling of 'Liaison' from faker.name.jobType().

fdsimms:
One of the possible return values for `faker.name.jobType()` was "Liason", a misspelling of "Liaison". I've fixed that in the lib/ files.

I only committed the four actual changes I made to those files, and not the files that resulted from building with `gulp`—I assumed that was the way to go, but let me know if that was wrong.


Marak:
Good catch!

Thank you for your contribution @fdsimms

PR #382, opened 2016-06-13
feature(creditCard) - finally here

zpiman:
**Intro**
I think generating random, but (sort-of) valid credit cards is very useful for testing online stores etc. For a recent project I required a large number of credit cards in random formats and so as always I looked to `faker.js`. However to my surprise, faker does not support credit cards yet! I started digging around and found that a while back in issue #169 and later in PR #258 someone made an attempt to implement it. However due to some licensing issues the PR died....

**What I did**
So I decided to give it a go and write it from scratch - no LICENSE required!!!!. I found that among to locals, there is already a scheme for the credit cards (used to be under `lib/locals/en/credit_card`). I decided to move it into finance keeping the hierarchy... `/lib/locals/en/finance/credit_card`.

I updated the schemes but realized that they use a sort of reverse RegExp notation where `[1-5]` generates a random number between 1 and 5 and `{1,4}` repeats the preceding character 1 to 4 times e.g. `"51[1-5]#{1,4}"`-> `"544###"` I implemented a simple helper method 'regexpStyleStringParse' that can parse it (it is not done though, I plan to also make it work with the rest of characters like in RegExp e.g. `[a-z], [abc], [a-zA-Z0-9]` <- _could be useful in other cases as well_ and implement some bracketing in the future perhaps so that things can be nested....)

I also wrote a `replaceCreditCardSymbols` helper method that helps to formalize the credit card number and replaces the `"L"` at the end with the Luhn checksum number. This required the implementation of the Luhn checksum algorithm which is described here https://en.wikipedia.org/wiki/Luhn_algorithm

Of course I also implemented all the tests as to make sure all works fine. This can however still be improved! I imagine writing a testing Regular expression for each format... as of now I only had the time to write it for visa cards and maestro :smiley: 

Hope you like it and enjoy! :beers:


zpiman:
Hmmmmm.... somethings not right....

Marak:
Marked for review, thank you.

Will get back to you soon.

alex051186:
Did you merge #382 to master ?

zpiman:
Its a shame, Looks like nothing has changed. Did I do all the work for nothing? I was hoping someone could check the code so that it can be merged... :cry:

alex051186:
@zpiman Your code very great and helpful, thank you so much, i'm here waiting for merge 👍

lib/helpers.js
```diff
@@ -1,3 +1,18 @@
+/** Define the String.prototype.repeat if it is not implemented.... See PR #382
+*/
+if(typeof String.prototype.repeat === "undefined") {
+  String.prototype.repeat = function(num) {
```
Marak: We can't prototype string.

Should move / remove this.

zpiman: This was only a fix for node < 0.1x. Back then, `String.prototype.repeat` did not exist. I am not sure anyone still uses that version of node. However I decided to rather create a dedicated helper function. It might not be optimized as the built in, but its not used that often...

zpiman:
Ehmmm, @Marak, do we need to support node 0.10 and 0.11? I don't know what is up with Travis but it is failing on those two versions. I have performed a number private tests on both 0.10 and 0.11 and they work for me.

lib/helpers.js
zpiman: This is the test passing on my machine, I added a console.log of the node version to the output...

zpiman:
There we go, through the act of printing out debug messages, the test suddenly magically works....

usebaz:
@Marak did you merge #382 to master ?

Marak:
I'm seeing sporadic errors for `finance.js creditCardNumber() returns a random credit card number`

Seems to be generating invalid number every so often. Investigating further.

test/finance.unit.js
```diff
@@ -222,11 +222,80 @@ describe('finance.js', function () {
...
         });
     });
 
+    describe("creditCardNumber()", function(){
+      var luhnFormula = require("./support/luhnCheck.js");
+      it("returns a random credit card number", function(){
+        var number = faker.finance.creditCardNumber();
+        number = number.replace(/\D/g,""); // remove formating
+        console.log("version:", process.version, number, number.length);
+        assert.ok(number.length >= 13 && number.length <= 19);
```
Marak: Problem here is that method sometimes returns value that is 20 long. This causes test to fail 5% of the time or so.

PR #386, opened 2016-07-04
test if root is defined

YannickBochatay:
root is undefined when the file is loaded as a module


Marak:
Can you be more specific?

YannickBochatay:
When you use a module bundler as webpack, the files are loaded as modules, and according to the [spec](http://exploringjs.com/es6/ch_modules.html) in a module "this" is undefined.
So if you check if the variable root is not undefined before trying to access its properties, it will work with module bundlers.

Marak:
This should have been resolved with https://github.com/Marak/faker.js/commit/02be0dc61e0ff84bdbe975d66b09c6231b25542c

Closing

PR #390, opened 2016-07-12
Fix output of faker.finance.amount.

### LINKED ISSUE #361: faker.finance.amount does not randomize the digits anymore
Hello, I was using faker 3.0.0 and everything was working perfectly. When I updated to 3.0.1 (and same with 3.1.0), faker.finance.amount now only provides "000" as digits.

Code: 
```
faker.finance.amount(58,60,4)
faker.finance.amount(58,60,4)
faker.finance.amount(58,60,4)
faker.finance.amount(58,60,4)
```

Pre-3.0.1:
```
59.4567
60.3901
58.7192
58.8835
```
3.0.1 - 3.1.0:
```
58.0000
59.0000
58.0000
60.0000
```

I don't think this is the expected behaviour.

jreina:
This simplifies the faker.finance.amount function a bit by avoiding arithmetic, and allows faker.random.number to do the heavy lifting with a precision operator. Fixes #361

### DIFF
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -88,9 +88,9 @@ var Finance = function (faker) {
       max = max || 1000;
       dec = dec || 2;
       symbol = symbol || '';
-      var randValue = faker.random.number({ max: max, min: min });
+      var randValue = faker.random.number({ max: max, min: min, precision: Math.pow(10, -dec) });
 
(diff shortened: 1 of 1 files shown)


Marak:
@jreina 

I think that is right.

Merged.

PR #392, opened 2016-07-24
Added comma replace for price method.

Zeludon:
E.g: 1000000 into 1,000,000.


Marak:
@Zeludon 

Thanks!

Could you add a test for this?

Zeludon:
I've never really written unit tests before, will this do?

PR #395, opened 2016-07-27
Added ethereum address faker and test

jackreichelt:


davesag:
👍 I could definitely use this.

lib/index.js
```diff
@@ -85,6 +85,9 @@ function Faker (opts) {
   var System = require('./system');
   self.system = new System(self);
 
+  var Ethereum = require('./ethereum');
```
Marak: Instead of being a top-level module, `Ethereum` should be sub-classed under `Finance` ( the same as bitcoin currently is )

Marak:
@jackreichelt 

This is good.

Only thing is that ethereum should be inside the `finance` module, instead of it's own top-level faker module.

If you can redo the request I will merge, if not I will try to make the change myself.

Thanks!

jackreichelt:
@Marak I'll update it tomorrow and resubmit this.

Marak:
Resolved with https://github.com/Marak/faker.js/pull/514

PR #397, opened 2016-08-04
Added some names

perstromswe:


lib/locales/sv/name/first_name_women.js
```diff
@@ -1,12 +1,29 @@
...
+  "Agneta",
+  "Anette",
+  "Anneli",
+  "Alexandra",
+  "Agnes",
+  "Anne",
+  "Alva",	
+  "Alma",	
+  "Angelica",	
+  "Ann-Marie"
```
jackreichelt: Comma after this one, mate.

Marak: Correct. The `Ann-Marie` name here seems to have been causing the build to fail for this PR.

Issue #404, opened 2016-08-30
helpers.shuffle on empty array causes an explosion


HowWeRollingham:
```
> console.log(faker.helpers.shuffle([])[0]);

<--- Last few GCs --->

   75912 ms: Mark-sweep 1120.9 (1255.0) -> 895.8 (1255.0) MB, 3681.1 / 0 ms (+ 507.9 ms in 1166 steps since start of marking, biggest step 3.5 ms) [allocation failure] [GC in old space requested].
   78653 ms: Mark-sweep 895.8 (1255.0) -> 869.2 (1255.0) MB, 2740.7 / 0 ms [allocation failure] [GC in old space requested].
   81698 ms: Mark-sweep 869.2 (1255.0) -> 869.1 (1255.0) MB, 3045.0 / 0 ms [last resort gc].
   85150 ms: Mark-sweep 869.1 (1255.0) -> 869.1 (1255.0) MB, 3452.2 / 0 ms [last resort gc].


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x3b8ab77c9e59 <JS Object>
    1: shuffle [/Users/ben/Projects/edwp/edwfp-data-service/node_modules/faker/lib/helpers.js:~85] [pc=0x236e86585d42] (this=0x1c8c0250ff41 <JS Object>,o=0x3773d1795aa9 <JS Array[0]>)
    2: /* anonymous */ [repl:1] [pc=0x236e8650614c] (this=0x3b8ab77e4dc9 <JS Global Object>)
    6: defaultEval [repl.js:272] [pc=0x236e865ac3a0] (this=0x37ecb4a66989 <an REPLServer with map 0x295dfe542111>,code=0...
```

Marak:
Resolved with #405

Marak:
Closing.

Should have been resolved a long time ago.

PR #405, opened 2016-08-30
Prevented a crash on running helpers.shuffle on an empty array

HowWeRollingham:
* Fixes bug #404


Marak:
@HowWeRollingham 

Good catch! Thanks.

Merged.

PR #412, opened 2016-09-16
Fix case where root would be undefined

rexxars:
In rare cases, I've found `this` to be undefined when trying to use this module. I think it's somehow related to webpack in eval mode, but can't quite figure it out. This PR stops it from crashing when trying to access properties on `undefined`.

### DIFF
--- a/vendor/password-generator.js
+++ b/vendor/password-generator.js
@@ -62,4 +62,4 @@
   }
 
   // Establish the root object, `window` in the browser, or `global` on the server.
-}(this));
+}(typeof window === 'undefined' ? global : window));


Marak:
@rexxars 

Can you provide any additional information on the error?

A reproducible use-case would be good.

I would just merge, but I'm not immediately sure if this will break other use-cases.

bjoerge:
This will happen if a bundler wraps the contents of the file in a function that has "use strict", e.g.:

```js
(function() {
"use strict"
// insert contents of bundled file here
})
```
This seems to be what webpack effectively does in eval mode:

image

Marak:
I think was resolved with 02be0dc61e0ff84bdbe975d66b09c6231b25542c

Issue #407, opened 2016-09-01
Spaces break `f.fake()`


paolo-tanium:
![image](https://cloud.githubusercontent.com/assets/21039701/18177780/ac0d99ce-7049-11e6-8f11-094a1060a078.png)

Marak:
It would be good if we had a better error message here or maybe just trim the whitespace by default.

PR #414, opened 2016-09-25
Fixing spaces in method

paolodm:
As a first time user of faker, I wrote something akin to the following:

``` js
faker.fake('{{ name.lastName }}, {{ name.firstName }} {{ name.suffix }}');
```

And it failed with the following error:

```
Error: Invalid module: name
```

It took me awhile to realize that the problem was due to the spaces between `{{` and `name`. 

It's not a big deal _once you know it_, but it does make the learning curve a bit harder for new users. Spaces in brackets seems totally reasonable, and we should handle this case. This pull request does that.

### DIFF
--- a/lib/fake.js
+++ b/lib/fake.js
@@ -32,23 +32,15 @@ function Fake (faker) {
       return res;
     }
 
-    // find first matching {{ and }}
-    var start = str.search('{{');
-    var end = str.search('}}');
+    var methodMatch = str.match(/{{(.*?)}}/);
 
     // if no {{ and }} is found, we are done
-    if (start === -1 && end === -1) {
+    if (!methodMatch) {
       return str;
     }
 
-    // console.log('attempting to parse', str);
-
-    // extract method name from between the {{ }} that we found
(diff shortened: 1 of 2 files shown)


Marak:
@paolodm -

Good catch!

Will review and merge.

Has this issue been previously reported?

paolodm:
I opened #407 a few weeks ago for this very issue.

Marak:
Nice. 

If you make a PR in the future, try to reference any potential open issues by number.

It helps track things as Github will automatically add an activity event the issue.

Marak:
On second thought, it seems to be we shouldn't allow users to enter ambiguous syntax into the method. It will only lead to more complicated code paths in the future.

It would be nice if the `faker.fake` at least gave back a more meaningful error message.

I'll take another look and see if I can include this in the next release.

Marak:
I don't think it's a good idea to introduce ambiguous syntax and additional code.

It's better developers stick to a single syntax and we keep the code paths easier to read.

Let me know if you have any other ideas.

PR #419, opened 2016-10-03
[api][dist] Add system methods, update travis.yml

cvega:
- Returns a random path with file name
- Returns a random directory path (no trailing slash)


cvega:
We can expand on this if you like (e.g. platforms(unix,win), trailing slash, more directories, etc..). I was also curious how you felt about a system.unit.js test using regular expressions to assert?

The directories are mostly top level from common unix file systems (linux, freebsd, OS X). We can go deeper. I was thinking I could grab directories from vanilla installs and populate lists for specific platforms.

cvega:
Added latest versions of node to travis too.

Marak:
Looks good, thank you @cvega

PR #421, opened 2016-10-04
#420 Added dataurl method

### LINKED ISSUE #420: feature : provide dataURL or images for offline use
When working offline images URL like the ones provided by `image.avatar` are not available.

Is it possible to provide dataURL or images from the faker package instead ?
maybe like that : 
`image.placeholder(width, height)`

sylvainbannier:
Adds `dataurl(width,height)` method that generates a dataurl embeded SVG image like this : 
generated image

fixes #420

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -192,7 +192,19 @@ var Image = function (faker) {
    */
   self.transport = function (width, height, randomize) {
     return faker.image.imageUrl(width, height, 'transport', randomize);
-  }  
+  };
+  /**
+   * dataurl
+   *
+   * @param {number} width
+   * @param {number} height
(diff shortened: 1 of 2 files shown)


lib/image.js
```diff
@@ -192,7 +192,19 @@ var Image = function (faker) {
...
-  }  
+  };
+  /**
+   * dataurl
+   *
+   * @param {number} width
+   * @param {number} height
+   * @method faker.image.dataurl
+   */
+  self.dataurl = function (width, height) {
```
Marak: I think this would be called `datauri` instead of `dataurl`?

Marak:
This was merged into `master` a few commits ago.

Thank you!

PR #423, opened 2016-10-17
Add ABA routing number generator

joaopaulofonseca:
This PR adds a generator for ABA routing numbers.

### DIFF
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -36,6 +36,27 @@ var Finance = function (faker) {
   }
 
   /**
+   * routingNumber
+   *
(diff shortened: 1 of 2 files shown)


Marak:
@joaopaulofonseca 

Looks OK.

Could you improve on the test slightly? Seems a bit broad, would be nice to have a regex to test against.

Marak:
Merged. Would be nice to have better tests for this, but it does seem to be working as expected.

Marak:
Thank you @joaopaulofonseca!

PR #427, opened 2016-10-24
add date format option with tests - update deps

kristianmandrup:
Please review. All tests pass locally. No more deprecation warnings :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
     "doc": "jsdoc -c conf.json -t ./node_modules/ink-docstrap/template -R README.md lib"
   },
   "devDependencies": {
-    "browserify": "5.11.1",
-    "gulp": "3.8.8",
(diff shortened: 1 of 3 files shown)


Marak:
Don't think accepting`moment` as a dependency is a good idea.

Seems outside the core functionality of `faker`.

package.json
```diff
@@ -17,8 +17,8 @@
     "doc": "jsdoc -c conf.json -t ./node_modules/ink-docstrap/template -R README.md lib"
   },
   "devDependencies": {
-    "browserify": "5.11.1",
-    "gulp": "3.8.8",
+    "browserify": "^13.1.1",
```
Marak: I wanted to include these changes, but since they were grouped with a separate set of commits ( date / time functionality ) I was unable to merge.

PR #432, opened 2016-11-01
Add .npmignore to reduce size of package

schovi:
In our company we use Docker containers for development. Once I checked size of Images and was amazed how 2 different Node application have 300 MB diff size. After some digging I found it is just in npm packages. One of then is this Faker.js

When checking size after npm install it give me this: 
```
du -hd0 node_modules/faker
74M    node_modules/faker
```

I decided to give it some time and check what can take so much I came with this. Adding `.npmignore` wont publish selected files to npm repository. It have many bonuses like saving network footprint to faster installation.

After adding `.npmignore` result is following:
```
du -hd0 node_modules/faker
9.3M    node_modules/faker
```

I tried to run tests on https://github.com/apiaryio/dredd which use this package and it works fine.

Of coure I dont know lot about this package, but I followed common sense and remove what deserves it :)

Give me hint if it is OK, should I add something or remove something more?

PS: there is how to test it https://docs.npmjs.com/misc/developers#before-publishing-make-sure-your-package-installs-and-works

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,7 @@
+.npm/
+build/
+doc/
+examples/
+meteor/
+test/
+logo.png


knownasilya:
Alternate approach is specifying the files you export in the `files` property in `package.json`.

schovi:
@knownasilya Is there any pros/cons in one and other approach?

thomaschaaf:
What is the process for this?

schovi:
@thomaschaaf Have no idea :(

Marak:
@schovi - I think we have to add this now to assist with enterprise support.

The problem is, I really liked the idea of having all the examples and builds included in the npm package. It can really assist novice users.

Not sure if there is anything else to do. Reducing the package size is a good idea all around.

knownasilya:
Or go to github directly and look at docs, site, examples there.

Marak:
@schovi @knownasilya 

Merged.

You are right. I generally try to always try to keep a few running examples in a package, but `faker` package is currently bundling all individual locale builds and examples ( a few times ).

Individual locale builds are taking up close to 70 megs ( as seen ). This is no good.

I'll see if I can sneak in a few of the very small node examples, but overall in the next release of `faker` we will see package size significantly reduced.

Thank you for your contributions. This is a really good update to the project for such a small delta.

PR #437, opened 2016-11-24
[api] Added `lorem.slug` method

lencse:
Returns a random slug for URLs


Marak:
@lencse - Why not! Merged.

Seems like a simple useful addition and it has tests.

Thank you for your contribution. I'll try using this method in one of my apps soon.

Will be included in the next release.

lencse:
Great, thanks for your feedback

PR #439, opened 2016-11-30
Add notes about seed.

NickHeiner:
This would have helped me!


Marak:
@NickHeiner -

I can't merge this PR since the change is to the auto-generated `README` file. The change needed to be made to the source file for the documentation.

I went ahead and just copy and pasted the change to the right place. Thank you for your contribution.

PR #440, opened 2016-11-30
Don't allow path seperators in generated filenames

mkornblum:
A number of the word lists contain "24/7" as a word, for example: https://github.com/Marak/faker.js/blob/e8231413cc38a3ce49d712823dadbffbd1849acc/lib/locales/en/company/bs_adjective.js#L17

This means that calling `faker.system.fileName()` or `faker.system.commonFileName()` could sometimes yield something like "some_words_24/7_yep.jpg". This implies a file called "7_yep.jpg" in a directory called "some_words_24", which is probably not what someone calling these methods wants.

### DIFF
--- a/lib/system.js
+++ b/lib/system.js
@@ -19,6 +19,7 @@ function System (faker) {
     str = str.replace(/\,/g, '_');
     str = str.replace(/\-/g, '_');
     str = str.replace(/\\/g, '_');
+    str = str.replace(/\//g, '_');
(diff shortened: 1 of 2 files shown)


Marak:
@mkornblum -

Very good catch.

I've merged this into `master`  and it will be included in the next release.

mkornblum:
happy to help! 👍 👍

Issue #430, opened 2016-10-27
Error on requiring faker library 


rollymaduk:
Hi,
Please any ideas on how to fix this error below
`node_modules\faker\vendor\password-generator.js:19Uncaught TypeError: Cannot read property 'localPasswordGeneratorLibraryName' of undefined(…)`
in my browser console when I try to load my page with faker.js library, I am using node and browserfy.

mbeaudru:
I am encountering the same problem and I can't figure out why I have this error.

Installed with ```yarn add faker``` and importing, this same error pops out.

Did you figured this out ?

rollymaduk:
Hi @mbeaudru , I fixed this by adding a null checker in the affected file on line 19.
``` js  
/// changed the line below from
localName = root.localPasswordGeneratorLibraryName || "generatePassword",
//to 
 localName = root && root.localPasswordGeneratorLibraryName || "generatePassword",
```
I have made a PR #445  that addresses this.

rollymaduk:
@mbeaudru  please let me know if it solved the problem so I can close this issue

canhkieu:
Same issues, I was changed to but still error/
 localName = root && root.localPasswordGeneratorLibraryName || "generatePassword",

rollymaduk:
@canhkieu  how did you change it? did you fork the project and use npm link to locally test it? if you did that, then ensure you also removed your global and project modules by either using npm uninstall <package> -g or npm uninstall <package> -D or npm uninstall <package> --save,
you can also post your error logs to help me get a better picture of what is happening!
thanks

Marak:
@rollymaduk -

Does this actually solve the issue? What is the cause of the issue?

Not sure why this is happening. I'll see if I can investigate further.

Marak:
I've opted to move the password generation logic to faker core. The extra browserifying type logic in `vendor/password-generator.js` was causing issues.

Should be resolved in `master`.

Also resolves #460

PR #445, opened 2016-12-11
fix for issue #430

rollymaduk:
added a null checker line to address issue #430


dennisfleischmann:
Hi, why is that bug still not fixed?
Please add a bugfix to that issue

BR
Dennis

Marak:
Should be resolved in `master` as per https://github.com/Marak/faker.js/issues/430

PR #446, opened 2017-01-01
Price fix

dugoh:
Small fix for commerce.price()

### DIFF
--- a/lib/commerce.js
+++ b/lib/commerce.js
@@ -44,7 +44,7 @@ var Commerce = function (faker) {
    * @param {string} symbol
    */
   self.price = function(min, max, dec, symbol) {
-      min = min || 0;
+      min = min || 1;
       max = max || 1000;
       dec = dec || 2;
       symbol = symbol || '';
--- a/test/commerce.unit.js
+++ b/test/commerce.unit.js
@@ -72,7 +72,7 @@ describe("commerce.js", function() {
       });
   });
 
-  describe("price(min, max, dec, symbol", function() {
+  describe("price(min, max, dec, symbol)", function() {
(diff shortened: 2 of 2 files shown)


Marak:
@dugoh -

What does this fix? Could you show us a small example of the current / expected behaviors?

Thanks!

lib/commerce.js
```diff
@@ -44,7 +44,7 @@ var Commerce = function (faker) {
    * @param {string} symbol
    */
   self.price = function(min, max, dec, symbol) {
-      min = min || 0;
+      min = min || 1;
```
alexpts: Why start value on default 1 better that 0?

dugoh: I don't know, but that is what the test expects. Without the fix the test fails about 1 in 1000 times.

dugoh:
@Marak 
If you look at line 2556 in https://travis-ci.org/Marak/faker.js/jobs/200239663 you see the output is missing a round bracket on the right.

Also the test for price has the assertion
        assert.equal((price > 0), true, "the amount should be greater than 0");
but the function defaults to a 0 minimum price causing the test to fail randomly about 1 in a 1000 times.

epayet:
I 👍 this, I just had a case where it was generating a price of 0 and breaking things, so I had to specify a minimum to fix it: `faker.commerce.price(1)` instead of just `faker.commerce.price()`.
But I would expect the function to have more than 0 as a minimum by default.

Plus, as @dugoh mentioned, the test specifically expect the price to be greater than 0, so it can fail randomly about 1 in a 1000 times.

Marak:
I see the issue with the tests now.

I guess it makes sense that a price should be `greater than 0` by default, although you could have a price of `0`.

PR #447, opened 2017-01-04
Method to generate a random CPF (Individual register of Brazilians)

cezarlz:


Marak:
@cezarlz - This is a good addition, but I'm not sure where to merge it into.

It feels like it should be part of the locality data, but `faker` doesn't really have any good ways for localities to export new methods.

I'm going to leave this PR open for now and see if we can find a good solution. It may require we refactor `faker` core a bit to allow for additions of custom methods for localities.

cristopher-rodrigues:
Hi @Marak some news about that?

glandre:
Hi @Marak and @cezarlz, this would be a very useful feature for Faker. A suggestion would be to use the library `cpf` (https://www.npmjs.com/package/cpf).

It generates CPF in a random way, and it follows the two-digit verification rule making sure that the CPF is valid.

It also provides the ability to generate invalid CPFs.

fnoquiq:
Please merge this

phaalonso:
Any notice of this module?

PR #453, opened 2017-01-15
Fix dec for amount

alexpts:
Now:
amount(10, 10, 0); // => 10.00

Expected:
amount(10, 10, 0); // => 10

-----

May be better in next major version change default value to 0.


Marak:
@alexpts - I see what happened here, nice catch!

What about the other the arguments near this section of code? Will they also be affected by `0` values?

alexpts:
@Marak I don`t check it, I think that Faker has tests for check it.

I don`t know need up major version because it change break old behavior or it will fix bug...

I can add research current tests for it method and add new tests for it change, if now there is not tests for this case.  Should I do it?

Marak:
@alexpts -

It would help if we could add tests for this functionality.

I think this will be OK and not break anything.

Thank you for your help. :-)

alexpts:
@Marak Ok, I will research current test and add new test for third param at weekend and send commit to current PR.

alexpts:
@Marak I added tests and fix similar for price method.

Also I not sure than amount and price must returns string type.
May be it need fix in other PR.

Marak:
@alexpts -

Merged.

I have a feeling we have similar issues sneak into in the code-base ( usage of `a || b` statements where `a` might be `0` ). This will always return false for `0`, which is probably not intended.

alexpts:
@Marak What do you think about returns Number type for simple price/amount without symbol?

PR #455, opened 2017-01-26
Add max, min for latitude and longitude

andresilvagomez:


Marak:
@andresilvagomez -

This appears to be good ( and correct ), but I'm not 100% certain without further review.

Can you ( or anyone else confirm ) this is a correct solution? Are there any cases where merging this Pull Request could break backwards compatibility? Is this still generating the same ranges as before?

Thank you!

andresilvagomez:
Hi @Marak, that works same, the new feature, has a specific function, the idea is limit the range of latitude and longitud, for example if you need create a coordinate in san francisco you send the correct params and that only generate coordinates in this area.  

Thanks !!!

lib/address.js
```diff
@@ -205,18 +205,26 @@ function Address (faker) {
    * latitude
    *
    * @method faker.address.latitude
+   * @param {Double} max default is 90
+   * @param {Double} min default is -90
    */
-  this.latitude = function () {
-      return (faker.random.number(180 * 10000) / 10000.0 - 90.0).toFixed(4);
+  this.latitude = function (max, min) {
+      max = max || 90
```
Marak: What if `max` value is `0`? This will always return `90`. Is that okay?

lib/address.js
```diff
@@ -205,18 +205,26 @@ function Address (faker) {
...
+      min = min || -90
```
Marak: What if `min` value is `0`? This will always return `-90`. Is that okay?

lib/address.js
```diff
@@ -205,18 +205,26 @@ function Address (faker) {
...
    *
    * @method faker.address.longitude
+   * @param {Double} max default is 180
+   * @param {Double} min default is -180
    */
-  this.longitude = function () {
-      return (faker.random.number(360 * 10000) / 10000.0 - 180.0).toFixed(4);
+  this.longitude = function (max, min) {
+      max = max || 180
+      min = min || -180
```
Marak: Same issues as above. A value of `0` for `min` and `max` will always return false.

Is that the intended behavior @andresilvagomez? Thank you.

andresilvagomez:
@Marak yes for all cases with 0 the real value is (90, -90) or (180, -180) because that latitude or longitude maybe is a error or mistake when developer are creating your fakers 👍

andresilvagomez:
@Marak talk me about that :) if you have more questions !

andresilvagomez:
@Marak 
Have you considered it?

PR #459, opened 2017-02-03
Update Faker.ObjectName references. These entries are First letter upper case, rather then lowercase

adeelshahid:


Marak:
`ReadMe` is auto-geneated file.

Can not merge.

PR #462, opened 2017-02-11
Docs: `seed()` is on `faker`, not `random`.

runeh:
The docs are using a non-extant method at the moment.


Marak:
@runeh -

Good catch!

Merged.

PR #465, opened 2017-02-14
docs: add instructions to install via npm

caroso1222:
Add npm install instructions.


caroso1222:
Why was this closed?

Marak:
Because the ReadMe is auto-generated through the build process.

askobara:
But README still doesn't contains any instructions to install the package. It's confusing a bit.

Issue #467, opened 2017-02-21
Company "Segment Name" Generation


sha1sum:
*Note:* This is a placeholder issue. I will be generating a pull request shortly.

In addition to the already-present data, I have found myself in need for randomized company "segment names", such as:

- international
- worldwide
- multi-continent
- continent
- country
- division
- region
- area

... and so on. I'll be generating a much larger list, and likely choosing the `company.segmentName` property, which can obviously be changed if preference swings toward another name.

Marak:
Sounds good!

sha1sum:
The related PR has been updated with changes from feedback.

PR #469, opened 2017-02-21
Company "Segment Names" (Administrative Sections)

sha1sum:
This includes the generation of a "segment name", also known as an "administrative division". When building a list of subdivisions within a company's organizational chart (divided management areas or jurisdictions), it sometimes is necessary to define the type of subdivision, which this would provide.

There are three main ways to call the `faker.company.segmentName()` function:

- `faker.company.segmentName()` - Give only generic names by default:
  - region
  - territory
  - area
  - division
  - branch
  - district
  - zone
  - sector
- `faker.company.segmentName(<level>)` - If building an organizational hierarchy, will ensure that a certain level of segment will appear, for example:
  - A level of `2` could generate "hemisphere", while
  - A level of `8` could generate "metropolitan area"
- `faker.company.segmentName(false)` - Include all levels when generating a segment name, irrespective of level.

For a list of "levels", see `lib/locales/en/company/segment_name.js`, where each array corresponds to a level, the first (index `0`) being the default generic terms if no level is provided as an argument. This is likely to be a non-comprehensive list, but I believe it to be a good start (searching for lists of these terms is a little abstract and obscure).

Addresses #467


Marak:
@sha1sum -

This is good!

Do you think we could add a bit more data for segments? Some of the levels don't have many options. We should ensure that every API endpoint of faker is going to return somewhat random data and add a few more segments per level.

sha1sum:
@Marak I'll see what I can come up with! It's a little hard to come up with these, but I'll do some more hunting, and I'm always open to suggestions 😄

Marak:
@sha1sum -

I see a few more we can add from https://en.wikipedia.org/wiki/Administrative_division for `Urban or rural regions:`?

Marak:
@sha1sum 

Any updates on this?

sha1sum:
@Marak sorry about this, it's been a while since I have been able to look over my open source requests! I have added a few more segment names, most of them from the `Urban or rural regions` section you suggested from the Wikipedia article. :)

sha1sum:
As this has been lingering for a long time, I am going to close this. If anyone else wants to continue this then feel free.

PR #477, opened 2017-02-28
adds optional `exclude` parameter to .words()

### LINKED ISSUE #449: [feature request] allow user to exclude certain words from .words()
I use `.words()` to simulate test data, and sometimes it comes back with words that cause a validation function to fail. It would be helpful if the user could optionally pass in an array of words to exclude from the output, so that any time those words came up they would be 'rerolled'. 

Happy to put in a PR on this too :)

Thanks!

martahj:
This resolves #449 by allowing the user to pass an optional second parameter to `.words`, an array of words to exclude from the output. When a word is selected by the`.word()` call, it will check to see if the word is part of the blacklist (which is first packaged into an object to avoid repeatedly iterating over an array), and if so will select another word until an allowed one is produced.

### DIFF
--- a/lib/lorem.js
+++ b/lib/lorem.js
@@ -23,11 +23,21 @@ var Lorem = function (faker) {
    * @method faker.lorem.words
    * @param {number} num number of words, defaults to 3
    */
-  self.words = function (num) {
+  self.words = function (num, excludeList) {
       if (typeof num == 'undefined') { num = 3; }
+      if (!Array.isArray(excludeList)) { excludeList = []; }
+      var exclude = excludeList.reduce(function (excluded, wordToExclude) {
+        excluded[wordToExclude] = true;
+        return excluded;
+      }, {});
       var words = [];
+      var getNextWord = function () {
+        var nextWord = faker.lorem.word();
+        if (exclude[nextWord]) { return getNextWord(); }
+        return nextWord;
+      };
       for (var i = 0; i < num; i++) {
-        words.push(faker.lorem.word());
+        words.push(getNextWord());
       }
       return words.join(' ');
   };
--- a/test/lorem.unit.js
+++ b/test/lorem.unit.js
@@ -32,6 +32,20 @@ describe("lorem.js", function () {
                 assert.equal(words.length, 7);
             });
         });
+
+        context("when 'exclude' param passed in'", function () {
+          it("excludes the requested words", function () {
+            var exclude = faker.definitions.lorem.words.slice(1);
+            assert.ok(Array.isArray(exclude));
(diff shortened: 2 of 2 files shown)


Marak:
I'm not 100% sure I like this.

The problem is that this pull request introduces the concept of uniqueness for API calls ( which has been discussed in other issues ). This type of functionality should probably be done library wide.

I can see that this PR is specifically for an *exclude* list of words, but it's the same kind of logic as *unique* values from API calls. It's essentially calling the faker API method in a loop until faker returns a satisfied value ( in this case, one that has been excluded ). Issue I have is that there is no guarantee that the faker method will ever complete. We also have no guarantee of the speed at which the API method will complete. I don't want to introduce any API semantics that are unpredictable or not configurable.

My point is that if we are going to introduce this kind of functionality we should do the following:

 - create separate faker method which can perform excludes / uniques on any API method
 - ensure this method has a time-out value using a timer to ensure max timeouts of calls
 - ensure this method has options for configuring min / max iterations ( to ensure performance )

Please feel free to iterate on this PR. I do want to see this kind of functionality added to faker, just in the right way.

martahj:
Hm yes I see what you're saying. In practice (while testing) the tests continued to run fairly quickly even though the test passes almost the entire array of `faker.definitions.lorem.words` as the array of words to exclude, but there is no guarantee that it will always run quickly - and if you pass the entire `faker.definitions.lorem.words` (or a superset of it) as the exclude list, you'll get an infinite loop.

Question about what you are thinking in terms of the last two bullet points - say the timeout is exceeded, or the maximum number of iterations is exceeded, without finding a unique/nonexcludeded value. What should be returned from the api?

Marak:
I've gone ahead and added the ability to generate unique results for any faker method.

Here is an example: https://github.com/Marak/faker.js/blob/master/examples/node/unique-values.js

I think the next step here is to provide an option for `Faker.unique` to pass in array of excluded words. I may also just alias this to a new `Faker.exclude` method which wraps `Faker.unique`. The bulk of the logic for excluding value is already handled with `unique`.

Marak:
@martahj -

I've reworked the concept here into a new module for [uniqueness](https://github.com/Marak/faker.js/blob/master/vendor/unique.js).

This API should now work in `master`

```js
var excludeArr = ['nesciunt', 'qui', 'dolor'];
var result = faker.unique(faker.lorem.words, [2], { exclude: excludeArr });
```

`faker.unique` takes three arguments. `method`, `arguments`, `options`. 

In `options` we can specify an array of `exclude` values to be excluded  from results. Implementation has time / loop count / predictable state.

Let me know if this would work for the feature you originally requested. Thank you for your contributions.

PR #478, opened 2017-02-28
Fix generation of float numbers

### LINKED ISSUE #401: random.number precision option needs improvement
using this code:

`var faker = require("Faker")
var options = {min: 0, max: 10, precision:1}
options.precision
faker.random.number(options)`

i would expect a number between 0 and 10 to be returned, but i just get NaN. looking at the code this is what i would expect to happen.
I am simply trying to understand the methods in order to generate a float, there doesnt appear to be much documentation around these methods.

Does anyone have any idea what i am doing wrong?

IvanGoncharov:
Fixes #401

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -47,8 +47,10 @@ function Random (faker, seed) {
       max += options.precision;
     }
 
-    var randomNumber = options.precision * Math.floor(
+    var randomNumber = Math.floor(
       mersenne.rand(max / options.precision, options.min / options.precision));
+    // Workaround problem in Float point arithmetics for e.g. 6681493 / 0.01
+    randomNumber = randomNumber / (1 / options.precision);
 
     return randomNumber;
 
--- a/test/random.unit.js
+++ b/test/random.unit.js
@@ -56,6 +56,14 @@ describe("random.js", function () {
 
     });
 
+    it("provides numbers with a with exact precision", function() {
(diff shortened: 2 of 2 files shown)


lib/random.js
```diff
@@ -47,8 +47,10 @@ function Random (faker, seed) {
       max += options.precision;
     }
 
-    var randomNumber = options.precision * Math.floor(
+    var randomNumber = Math.floor(
       mersenne.rand(max / options.precision, options.min / options.precision));
+    // Workaround problem in Float point arithmetics for e.g. 6681493 / 0.01
+    randomNumber /= 1 / options.precision;
```
Marak: Could you possibly expand this out to make it easier to read?

I'm not actually certain what this is doing.

Thanks!

Marak:
@IvanGoncharov -

Looks good. Could you adjust the code to be a bit more easier to understand? I made a comment.

IvanGoncharov:
@Marak Done

Marak:
Thanks @IvanGoncharov, very nice!

IvanGoncharov:
@Marak My pleasure. Can you please publish npm package with this fix?

Marak:
I'll do a patch version in an hour or two.

IvanGoncharov:
@Marak Thank you.

PR #481, opened 2017-03-03
Added OpenID Standard Claims format for en_GB

mikefrancis:


mikefrancis:
@Marak Any word on whether this is good to merge?

Marak:
Seems simple enough, but could break the API since the other formats don't have the `+` symbol included?

PR #485, opened 2017-03-07
Fix typo in city names

atorkhov:


Marak:
Good catch @atorkhov 

Merged.

PR #488, opened 2017-03-08
Add 'setLocale' method to solve problem with ES6 import

IvanGoncharov:
If you do `import * as faker from 'faker'` you can't do `faker.locale = 'uk';` since it produce runtime errors. 

ES6 forbids you to modify properties of top-level important objects and same is also true for TypeScrips.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -238,6 +238,8 @@ Setting a new locale is simple:
 
 ```js
 // sets locale to de
+faker.setLocale("de");
+// or
 faker.locale = "de";
 ```
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -142,6 +142,10 @@ function Faker (opts) {
 
 };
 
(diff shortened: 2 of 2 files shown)


IvanGoncharov:
@Marak can you please review this change. It's only a few line of code but big change for developers who using your library in ES6/TS/Flow projects.

Marak:
I guess that makes sense.

Will review and add to next version.

IvanGoncharov:
@Marak Can you please merge this PR and release a new version?

Marak:
@IvanGoncharov 

Could you add test for this?

Marak:
It seemed like we should merge this ASAP, so I have.

Please try to add tests later if you can.

IvanGoncharov:
@Marak Can you please release new NPM package with this change included?

PR #494, opened 2017-04-03
Remove emdash from pl's first_name.js

tullmann:
I think this em-dash was separating the male and female names?  Its definitely not a good first name to suggest.  (I noticed when faker.internet.domainWord() returned this em-dash.)


Marak:
That is odd. Good catch.

Merged.

Thank you.

PR #502, opened 2017-05-01
Use `src="` instead of `src = "`

adriaanvanrossum:


Marak:
Documentation changes should be done in the source files. 

ReadMe is auto-generated.

https://github.com/Marak/faker.js/blob/master/build/src/docs.md

adriaanvanrossum:
Okay bye

PR #511, opened 2017-06-08
Add faker.markdown

mrtc0:
Added `faker.markdown`

### DIFF
--- a/lib/markdown.js
+++ b/lib/markdown.js
@@ -1,0 +1,114 @@
+
+/**
+ *
+ * @namespace faker.markdown
+ */
+var Markdown = function (faker) {
+  var self = this;
+  var Helpers = faker.helpers;
+
+  /**
+   * header
+   *
+   * @method faker.markdown.header
+   * @param {number} num number of '#', defaults to 1
+   */
+  self.header = function (num) {
+    if (typeof num == 'undefined') { num = 1; }
+    var head = new Array(num + 1).join("#")
+    return [head, faker.lorem.word()].join(' ')
+  }
+
+  /**
+   * emphasis
+   *
+   * @method faker.markdown.emphasis
+   * @param {string} type emphasis
+   */
+  self.emphasis = function (type) {
+    var types = [
+      '_',
+      '~',
+      '*',
+      '**'
+    ];
+    var words = faker.lorem.words(3).split(' ')
+    var position = faker.random.number(words.length);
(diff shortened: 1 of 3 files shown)


Marak:
This is pretty neat.

I might want to see this added to `lorem` module instead of it's own module though.

mrtc0:
Thanks.

Why should I add it to the `lorem` module?

Marak:
I was thinking `lorem`, since this functionality is wrapping existing `lorem` api methods. That would require we namespace to `lorem.markdown.foo`, which goes against how that rest of faker API works. Probably not a good idea to do that actually.

We could add `markdown` as a top-level module ( like this PR ), but the current functionality doesn't seem important enough to warrant it's own top-level module.

**Would you be willing to add a bit more functionality to this PR?**

I was thinking we could add a few aggregate methods for generating types of fake example markdown documents, like a fake `ReadMe.md` document and maybe a fake blog post. We also may want more advanced options for the current markdown methods we have.

So, something like: `faker.markdown.readme` and `faker.markdown.blogpost`

for inspiration look at:

https://jaspervdj.be/lorem-markdownum/
https://github.com/jaspervdj/lorem-markdownum

mrtc0:
I see, it is a good idea！

lib/markdown.js
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @method faker.markdown.unorderdList
+   * @param {number} num of list
+   */
+  self.unorderedList = function (num) {
+    if (typeof num == 'undefined') { num = 3 }
+    var words = faker.lorem.words(num).split(' ');
+    var list = [];
+    words.forEach(function (word, index) {
+      list.push(`* ${word}`)
```
Marak: Can you remove this template literal? It's going to break older javascript versions.

mrtc0: OK, I removed it.

Ethaan:
its this PR gonna be merged soon? @mrtc0 can i take what it needs to make this happend?

Marak:
Not sure why, but I just re-ran the build on Travis and it's still failing.

With the template literals removed it should be working?

I'll try to find the time to merge this locally for testing. Does anyone know if we missed a step here for Travis to pickup the modified commits?

Would like to get this merged in for the next release.

leesiongchan:
Any chance to get this merged?

Marak:
Tracking #919

sebacruz:
Any news on this?

PR #513, opened 2017-06-17
Update Random to create a per-instance copy of the RNG

chris-chambers:
Before this change, it appeared that multiple, independent Faker
instances were legal.  But, when multiple instances were created, their
shared random number generators would overwrite each others' states.

After this change, each Random instance (and therefore each Faker
instance) has its own copy of the random number generator, ensuring that
calls to one Faker instance do not affect the stream of values generated
by another.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -5,15 +5,27 @@ var mersenne = require('../vendor/mersenne');
  * @namespace faker.random
  */
 function Random (faker, seed) {
+  var gen = new mersenne.MersenneTwister19937();
+
   // Use a user provided seed if it exists
   if (seed) {
     if (Array.isArray(seed) && seed.length) {
-      mersenne.seed_array(seed);
+      gen.init_by_array(seed);
     }
     else {
-      mersenne.seed(seed);
+      gen.init_genrand(seed);
     }
   }
+
+  // Copied from the marsenne package for compatibility.
+  function rand(max, min) {
+    if (max === undefined) {
+      min = 0;
+      max = 32768;
+    }
+    return Math.floor(gen.genrand_real2() * (max - min) + min);
+  }
+
   /**
    * returns a single random number based on a max number or range
    *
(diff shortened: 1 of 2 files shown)


chris-chambers:
@Marak, sorry to ping you directly, but I wasn't sure how to advance this PR otherwise.  What steps do I need to take to get this ready to merge?

Marak:
Seems important.

Merged with 8a33cf0e7a57f8fe766034706d0230802788d1ff

Thanks @chris-chambers

Marak:
@chris-chambers 

Did you actually try testing this out with real-world usage?

After merging this change, faker is now returning the same exact results every time you run it. For example, now if I call `faker.name.findName()` multiple times in a row, it will always return the same value. Obviously this isn't the correct behavior. I'm reverting this commit for now.

Please let me know if you can provide a valid fix.

chris-chambers:
@Marak, I don't see the same behavior on my `random-independent-sequences` branch.  I am using these changes in a real-world app, but in my app, I seed the RNG explicitly each time, so that could be a cause of behavioral difference.

To be perfectly clear, I did:

```
git checkout random-independent-sequences
node
const faker = require('.');
faker.name.findName();
faker.name.findName();
faker.name.findName();
```

And the output was:
```
'Susanna Skiles'
'Tate Erdman V'
'Kasey Gottlieb'
```

chris-chambers:
Ah.  I see the trouble.  If I repeat those steps, I get the same values.  I have a patch I'm testing that should repair this issue.  Just need a test to cover the case.  Apologies for the breakage.

Marak:
Thank you for reviewing this.

I'd like to still merge your change, but I must make sure API doesn't break for others.

I think part of the problem is that test coverage didn't catch this. Please take your time to improve tests, thank you again.

chris-chambers:
@Marak, I've pushed a change and test that should correct the issue.

Marak:
That looks better @chris-chambers. 

I will do a more detailed review and should be including this in the next release.

chris-chambers:
Great!  Thank you!

Gregjarvez:
@Marak  Does this PR require any more work to be merged? I am happy to do any further work to get this merged. This is currently a blocker for my team and I.

chris-chambers:
This appears to be abandoned, feel free to use the code if it still makes sense.

Marak:
Tracking #918

PR #515, opened 2017-06-25
Fix spelling error

macsj200:
We want possessive its, not contraction it's (it is).


Marak:
ReadMe is auto-generated document. Need to make these kind of changes to the documentation source files.

Issue #461, opened 2017-02-09
Extra / duplicate `supplemental.js` files?


Marak:
It seems we have some un-needed duplication of lorem data for some locales in `supplemental.js` files.

Will investigate further. Could potentially be able to remove a few thousand lines of locality data!

Marak:
Resolved with #516

PR #516, opened 2017-07-03
Cleanup duplicate lorem/supplemental.js locale files and refs

thewhodidthis:
#461 I understand this is probably a superficial fix of minor importance, but since this is my first attempt at helping out, I thought I'd start somewhere safe. I believe people may have just copied over whole folder contents from ```locales/en``` when setting out to add the translations that produced the duplicates.  Before "purging" I ran,
```find . -name 'supplemental.js'```
to list the files in question,
```find . -name 'supplemental.js' -exec diff -q ./lib/locales/en/lorem/supplemental.js {} \;```
to track potential differences from English, only Latvian was a match, then
```grep -R 'supplemental' lib/*```
to check for references to affected files, x


Marak:
Thanks @thewhodidthis. 

This needs to be reviewed a bit since it's cross cutting, but it should get merged.

thewhodidthis:
KO, thanks, x

Marak:
Merged, thanks @thewhodidthis

PR #519, opened 2017-07-19
Update Build for correct street address in locales/de

jkimmeyer:
I updated the functions in the Build, because it's not possible to overwrite the Javascript bind-Method. Please fix it this way, or give another possibility to overwrite the generation of the street address.

### DIFF
--- a/lib/locales/fr/address/street_prefix.js
+++ b/lib/locales/fr/address/street_prefix.js
@@ -1,5 +1,6 @@
 module["exports"] = [
-  "Allée, Voie",
+  "Allée",
+  "Voie",
   "Rue",
   "Avenue",
   "Boulevard",


Marak:
What was the original issue @jkimmeyer?

I can't merge changes to the build file ( it's automatically generated )

jkimmeyer:
@Marak I can't generate a german street address. If i try to, an english(default) address is generated. But there is data for the street roots but not the correct syntax. Instead of the street_name the build should take the street_root and the street_root has to change the order with the number. 

I tried to overwrite the method in the locale file(none of the build), but it is not possible to overwrite the method.

Marak:
I can't see the code changes anymore, the PR went blank. :-\

It's of course possible to override any method in faker. Some of the locale data is not fully connected and it falls back to english locality for default.

Additional information would be appreciated. Thanks.

PR #530, opened 2017-08-25
[api] Add method `faker.random.float`

yjimk:
- Extends the `faker.random.number` float to automatically return a 0.01 precision float
- First parameter is set as the precision value

I've decided to create a PR for this feature because I was mocking float values here and there and found that setting with `faker.random.number({ precision: 0.01 })` was a bit verbose. Further to that, when I started using faker I didn't know how to create floats, took me a while to find out how I could do it. Hoping this method makes it easier and clearer from the start.

Love your work, it's been immensely helpful! Especially like some of the product names it generates, "Rustic Plastic Shirt" indeed. :)

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -57,6 +57,26 @@ function Random (faker, seed) {
   }
 
   /**
+   * returns a single random floating-point number based on a max number or range
+   *
+   * @method faker.random.float
+   * @param {mixed} options
+   */
+  this.float = function (options) {
+      if (typeof options === "number") {
+        options = {
+          precision: options
+        };
+      }
+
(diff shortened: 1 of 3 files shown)


lib/random.js
```diff
@@ -57,6 +57,26 @@ function Random (faker, seed) {
...
+  this.float = function (options) {
+      if (typeof options === "number") {
+        options = {
+          precision: options
+        };
+      }
+
+      options = options || {};
+
+      return this.number(Object.assign({}, options, {
```
Marak: Why `Object.assign` here?

PR #533, opened 2017-09-01
Fix spacing errors which prevented Markdown from displaying correctly

JasonMFry:


Marak:

PR #535, opened 2017-09-07
Update Australian postcode ranges

davidjb:
Ranges also include 1000-1999 and 8000-9999, see https://en.wikipedia.org/wiki/Postcodes_in_Australia#Allocation


Marak:
Seems correct.

Thank you @davidjb.

Merged.

PR #538, opened 2017-09-08
Add test for setLocale function added in #488

IvanGoncharov:
> Could you add test for this?

@Marak Done ✅


Marak:
Thank you @IvanGoncharov

Issue #537, opened 2017-09-08
Random alpha string?


dnd:
Apologies if there's someplace better to ask this, but is there not a way to generate a random string of alpha characters up to a certain length, or min/max length? `Random` has `numeric`, and `alphanumeric`, but not just `alpha`. Is this accomplished some other way?

Marak:
I suppose it would be easy to add `alpha` or `string` method to `Random` module.

You could also try running: `console.log(faker.helpers.replaceSymbols('?????'))`

`?` should replace with alpha.

aesyondu:
### replaces numbers.

PR #540, opened 2017-09-09
add `faker.random.alpha` method

dnd:
This PR implements random alpha string capability mentioned in #537

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -257,6 +257,40 @@ function Random (faker, seed) {
     return "0x"+wholeString;
   };
 
+  /**
+   * alpha
+   *
+   * @method faker.random.alpha
+   * @param {mixed} options defaults to 1
+   * @param {number} options.min minimum length defaults to 1
+   * @param {number} options.max maximum length defaults to 99
+   */
+  this.alpha = function alpha(options) {
+    if (typeof options === "undefined") {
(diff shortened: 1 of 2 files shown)


dnd:
@Marak I just checked on this PR and saw that it now has conflicts due to commit b198dda5d7a69d19f837aadd35c2ba8efc97a6ed. It looks like it's because my code still has the "independent" tests. 

Do you want to clean that up quickly, or do you want me to rebase, and force push the updates?

dnd:
@Marak is there anything that I need to do for this PR to get it accepted?

lib/random.js
```diff
@@ -257,6 +257,40 @@ function Random (faker, seed) {
...
+        options.max = 99;
+      }
+
+      count = faker.random.number(options);
+    }
+
+
+    var wholeString = "";
+    for(var i = 0; i < count; i++) {
+      wholeString += faker.random.arrayElement(["a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"]);
```
oudirichi: I know their is some case like this, but I suggest to define the array before the loop, this way the array is not defined every time

PR #543, opened 2017-09-13
The Indian postcodes are always numeric

kuldipem:


Marak:
Good catch.

Based on https://en.wikipedia.org/wiki/Postal_Index_Number, I believe you are correct.

Merged. Thank you for your contribution.

PR #551, opened 2017-09-27
Readme.md: Add note about `faker.seed(0)`

malthejorgensen:
I believe a lot of people will (like me) simply put `faker.seed(0)` in their test code.
However that won't work, so I think it makes sense to just add small note about that.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -315,6 +315,8 @@ var secondRandom = faker.random.number();
 console.log(firstRandom === secondRandom);
 ```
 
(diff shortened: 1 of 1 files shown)


backspace:
Thanks for documenting this, I ended up losing a lot of time to `0` not working 😢

Marak:
If `faker.seed()` with a value of `0` doesn't do anything it should throw an error or warning message.

Marak:

Issue #506, opened 2017-05-08
Request: Vehicle Data (VIN)


LvZeeland:
I'm building an app with vehicles and I was surprised faker does not offer any data for cars.
It would be a nice addition if requesting brands and realistic VINs is possible.

focusaurus:
Maintainers, if we submit a PR, would you consider reviewing and merging if acceptable? If so, @LvZeeland I'd be available to work with you on a PR for this data type.

Marak:
Sounds good. PRs welcomed.

pkuczynski:
This is done in #555 but not published yet (see #760) :(

PR #555, opened 2017-10-10
add Vehicle Namespace

tylerreichle:
# Vehicle Generator

- vehicle: returns random make/model combination
- manufacturer: random manufacturer
- model: random vehicle model
- type: random vehicle type
- fuel: random fuel type
- vin: random valid vin number
- color: random color

# random.alpha
 - generate random alpha characters.
 - `alpha()` returns single lowercase letter
 - `alpha({upcase: true})` returns single uppercase letter
 - `alpha({count: 5, upcase: true})` returns five uppercase letters

includes unit tests for all functions


tylerreichle:
#506

Marak:
@tylerreichle -

This is starting to look good. I notice at least one small code error, but overall I like.

Let me know when you are finished and I will review and merge.

lib/vehicle.js
```diff
@@ -0,0 +1,109 @@
...
+
+  self.vehicle.schema = {
+    "description": "Generates a random vehicle.",
+    "sampleResults": ["BMW Explorer", "Ford Camry", "Lamborghini Ranchero"]
+  };
+
+  /**
+   * manufacturer
+   *
+   * @method faker.database.manufacturer
```
Marak: I think this may be a typo?

tylerreichle: Wow we must have noticed the typo at the same time! I just corrected 😄

lib/locales/en/vehicle/vin.js
```diff
@@ -0,0 +1,32 @@
+module["exports"] = [
+  "YV1MH682762184654",
```
Marak: Are these all legitimate VIN numbers?

I tried looking up a few and they appear to be returning information for real cars.

If so, that could cause issue to owners of vehicle somehow.

tylerreichle: These came from an online vin generator that was using regex to return valid vins. I assumed users wanting to generate random vins would have a similar check to ensure they are valid.

Wonder how we can deal with generating valid vins without returning one in use?

tylerreichle: I could update to have vins returned from the included helper regexpStyleStringParse. That way no real vins will be hardcoded

Marak: It would be better I think if we could port the regex / code for generating VINs and place that in the API method.

I'm OK with generating valid VINs from algorithm. A pre-defined list seems bad because we would be singling out specific vehicles ( which could end up bad ).

Marak: That sounds good.

Let's see if we can remove the list of hard-coded VINs in favor of algo.

Would be nice if we could also have simple test, perhaps cross-reference documentation in comments for format of VIN specification.

Thanks!

tylerreichle: Yeah, I definitely see the problem now that you point it out! I will make those changes and let you know when ready for review

tylerreichle:
@Marak I updated vin to return using algorithm instead of hardcoded options. I also added random.alpha function for character generation in vin.

I also ended up using random.alpha in new airport namespace PR I just raised 👍

tilomitra:
Can we go ahead and merge this? It's just sitting here for 9 months.

Marak:
Marked for inclusion in next release `v5`

Marak:
Merged.

Thank you for your contributions @tylerreichle 

Closing.

pkuczynski:
Still not published? See #760

PR #562, opened 2017-10-21
Add Star Wars namespace

tylerreichle:
# Star Wars Generator

- `faker.starWars.character()` returns character
- `faker.starWars.droid()` returns droid
- `faker.starWars.planet()` returns planet
- `faker.starWars.species()` returns species
- `faker.starWars.vehicle()` returns vehicle/ship
- `faker.starWars.quote()` returns popular quote
- `faker.starWars.wookiee()` returns Wookiee word

includes unit tests for all functions
branched from add-airport

I enjoy having fun and interesting seeds and thought others would too 😄


Marak:
Not sure I'm willing to merge this PR. What value would it provide to 90% of the Faker users? I'm not certain it's worth the additional lines of code.

It's also a bit hard to review since it's grouped with all your other commits. I can't see what is new here versus what is part of the other commits.

tylerreichle:
That's understandable. I just noticed that the ruby faker has a lot of silly generators so thought it must be popular to have some fun ones.

I enjoy the silly names and locations in my seed data but guess that's not for everyone 👍

Marak:
Part of the problem is that I can't evaluate the size of the PR or what is changed since it's grouped with your other commits.

tylerreichle:
Yeah, that's my fault I assumed once the other PRs were merged the diff would be much less. I don't mind resubmitting from a clean branch if you thought it was worth checking out

Marak:
I think you are correct. 

Let me catch up on the other PRs and see if the the auto-diff corrects.

I've been trying to work out of a good general solution for allowing as much data into faker.js as possible without letting the library bloat too much in size.

tylerreichle:
Sounds good. I understand the want to cut down on the fluff.

Glad to be helping out though! Saw a few more open issues I will look into soon

MarcelMalik:
You also add the hacker namespace. Why not Star Wars?
For example one of my customers named there sample data after Star Wars planets.
I like the idea :)

Marak:
Tracking #919

Issue #496, opened 2017-04-04
Company.bs uses the wrong word order.


analytik:
Tested in faker 3.1.0 and 4.1.0.

Output I get: 
'plug-and-play engineer deliverables'
'global expedite convergence'
'leading-edge exploit schemas'

Expected outcome:
'engineer plug-and-play deliverables'
'expedite global convergence'
'exploit leading-edge schemas'

Marak:
Could you open a Pull Request to fix this?

PR #563, opened 2017-10-21
re-order of the company.bs adjective and buzz (#496)

mbehzad:
- re-orderes adjective and buzz in the company.bs template literal for a correct gramatics as sugessted in the issue #496,
- re-orders the statements in the unit test to reflect the same order as in the function for an easier read.


Marak:
Thanks @mbehzad! 

Merged.

Issue #457, opened 2017-01-31
Generate Fake Cardinal and Ordinal Directions?


jongunter:
I have an application that needs cardinal (N/S/E/W) and ordinal (NE/SE/SW/NW/) directions in its fake data. I was surprised that `faker.js` doesn't have these (since this library literally has everything!), and I ended up having to implement it myself. 

Just an idea for a feature to add down the road...

Marak:
@jongunter -

Seems like a good addition.

I'll merge any Pull Requests which add this functionality.

PR #564, opened 2017-10-22
add directions to Address namespace

tylerreichle:
# Directions

- `faker.address.direction()` returns random direction
- `faker.address.cardinalDirection()` returns cardinal direction (N/E/S/W)
- `faker.address.ordinalDirection()` returns ordinal direction (NE/NW/SE/SW)

includes unit tests

Addresses issue #457

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -233,12 +233,12 @@ function Address (faker) {
    * @method faker.address.direction
    */
   this.direction = function () {
-    return faker.random.arrayElement(faker.definitions.address.direction)
+    return faker.random.arrayElement(faker.definitions.address.direction);
   }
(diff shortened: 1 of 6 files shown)


Marak:
@tylerreichle -

Looking good!

Should we also add directions as whole words? `north` , `south`, `east`, `west`. I think this would help with generating addresses.

Thank you for your many contributions.

tylerreichle:
Glad to be helping out!

I updated the functions to return directions as full words unless optional `useAbbr` boolean is passed in as true. Let me know what you think about this option

Marak:
Looks good!

Merged.

Thank you.

PR #572, opened 2017-11-06
Fix pt_BR city suffixes

dserodio:
These are concatenated to cityPrefix without a space separator,
the example suffixes (which are full words) need to begin with a
space. I also added a `lândia` suffix which shouldn't begin with
a space.


Marak:
I believe the changes required for this might need to be done using the formatting strings we have per locality instead of changing the data.

Closing for now, but willing to re-open if we can.

PR #576, opened 2017-11-09
Add usage example into the README

ondrejsevcik:
Let's start to put usage examples in the readme. Sometimes it's not that obvious how to use that method and such a small example makes developer life much easier.


Marak:

Issue #575, opened 2017-11-08
Browser support through npm


caseywatts:
It looks like we got rid of the `builds/builds/faker.js` file in the published npm package, [here](https://github.com/Marak/faker.js/pull/432).

I'm working in a project that was depending on that one file. We usually wrap the browser-exported version (like `faker.js`). I'm not sure how else we can update to 4.0 without this built file being present:
https://github.com/samselikoff/ember-cli-mirage/issues/1037
https://github.com/samselikoff/ember-cli-mirage/pull/1062

Could we include just `builds/builds/faker.js` into the build? If we used the `package.json` feature `files`? (like mentioned [here](https://github.com/Marak/faker.js/pull/432#issuecomment-259269892))

At a glance, I think we would whitelist these folders/files:
```
lib
locale
vendor
index.js
builds/builds/faker.js
builds/builds/faker.min.js
MIT-LICENSE.txt
```

and maybe also
```
bower.json
package.json
package.js
```

Does that seem right?

Here's the full list of what's currently published to NPM:
![image](https://user-images.githubusercontent.com/273653/32568772-3ef311fe-c48d-11e7-8b46-9abfa81728f6.png)

caseywatts:
cc @schovi @knownasilya @Marak :)

btw - one reason I want to update to 4.0 is because the npm package filesize is so dramatically smaller! I'm totally onboard with the PR that did that ✨

Marak:
I think I see what happened.

@caseywatts - I'll see if we can re-include the main faker.js browser build file.

caseywatts:
I took a stab at it! :)

https://github.com/Marak/faker.js/pull/584

PR #584, opened 2017-11-30
don't npmignore build/build/faker.js 

caseywatts:
don't npmignore build/build/faker.js or build/build/faker.min.js
even though we ignore the rest of build/*

I tested it locally using `npm pack`

Fixes https://github.com/Marak/faker.js/issues/575

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,8 @@
 .npm/
-build/
+build/*
+build/build/*
+!build/build/faker.js
+!build/build/faker.min.js
 doc/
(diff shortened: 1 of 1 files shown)


Marak:
Thank you @caseywatts!

caseywatts:
@marak could we make a release that contains this sometime soon? If I can help with that let me know~

So excited to use 4.0 🤓

kellyselden:
Just hit these missing build files. A release would be appreciated.

Issue #466, opened 2017-02-14
Uniqueness


jaredpalmer:
Current experiencing issues with faker not generating unique enough values. This results in a false negative test result. Our workaround is to concatenate or prepend another faker function to increase randomness.

Notoriously not-unique-enough faker funcs for running integration tests with:
- `faker.internet.ip()`
- `faker.internet.email()`

Marak:
@jaredpalmer -

I can confirm conflicting values for both `faker.internet.ip()` and `faker.internet.email()` when running the attached code.

It seems on `master` we are seeing about 3 IP conflicts per 100,000 and 700 email conflicts per 10,000.

We should for sure fix this for email address. I should review other faker methods that may also require more entropy. RE: #245

```js

var faker = require('../../index');
faker.locale = "en";


var ips = {};
var conflicts = 0;
for (var i = 0; i < 100000; i++) {
  var ip = faker.internet.ip();
  if (typeof ips[ip] === 'undefined') {
    //console.log('set ip', ip)
    ips[ip] = ip;
  } else {
    conflicts++;
    console.log('CONFLICTS', ip)
  }
}
// 0 at 10,000
// 3 at 100,000


//console.log(faker.image.image())
var emails = {};
var conflicts = 0;
for (var i = 0; i < 100000; i++) {
  var email = faker.internet.email();
  if (typeof emails[email] === 'undefined') {
    //console.log('set email', email)
    emails[email] = email;
  } else {
    conflicts++;
    console.log('CONFLICTS', email)
  }
}
console.log('total conflicts', conflicts)

// 3 - 7 at 10,000
// 652 at 100,000
```

Marak:
I've looked at the code for generating emails / usernames and I'm not 100% certain why we are seeing conflicts at this scale.

https://github.com/Marak/faker.js/blob/master/lib/internet.js#L69

We should be generating enough uniqueness here to not be seeing conflicts with such small datasets? Can anyone confirm the math based on the code conditionals and first / last name data?

If anything, we may have to implement a way for faker to remember previously generated results and then re-generate until a new unique is found? I feel like we can improve the randomness without having to resort to that.

Marak:
RE: https://github.com/Marak/faker.js/pull/477

agm1984:
[...]

PR #596, opened 2018-01-12
Reset startTime appropriately

richchurcher:
Note #466 and [my comment on times](https://github.com/Marak/faker.js/issues/466#issuecomment-357140244). Refreshes `startTime` at 0 loop depth, preventing scripts which last longer than 5 seconds from throwing.

BTW I needed to use a fork of `gulp-gh-pages` because [this PR](https://github.com/shinnn/gulp-gh-pages/pull/117) isn't merged, which fixes [this issue](https://github.com/shinnn/gulp-gh-pages/issues/116), which is currently breaking your gulp build :smile:

### DIFF
--- a/vendor/unique.js
+++ b/vendor/unique.js
@@ -18,7 +18,7 @@ var maxTime = 5000;
 var maxRetries = 50;
 
 // time the script started
-var startTime = new Date().getTime();
+var startTime = null;
 
 // current iteration or retries of unique.exec ( current loop depth )
 var currentIterations = 0;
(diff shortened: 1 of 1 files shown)


Marak:
@richchurcher -

I ended up merging this, but it's still not 100% correct.

There were some other issues in `unique` module relating to scope of certain class variables being declared at the module level ( like `startTime` ).

I've done a basic fix but I think more testing will be required. If you have any time it would help if we could add more unit tests to cover this. I'll try to take a look at this again before release.

richchurcher:
Cool! I'll try to put that on my list of things to pay attention to :smile:

Issue #400, opened 2016-08-17
Zipcode and state don't match


si-mikey:
Is there a way to make faker.address.zipCode match a state, eg. 
faker.address.zipCode('NY') and it will generate ny zip codes?

nikmash:
I'm curious about this too. Anyone know if we can generate state specific zip codes?

si-mikey:
@nikmash  from the looks of it, it seems just to be random method specific data.

nikmash:
@si-mikey got it. Not ideal for testing with systems that do geocode resolution.

Marak:
@si-mikey @nikmash 

This seems like a useful feature to have.

I don't know exactly how, but it seems like there is a known algorithm for determining zipcode min max ranges by state?

Check this out: http://www.structnet.com/instructions/zip_min_max_by_state.html

If anyone could turn this into code, or find an existing JavaScript solution which does, I'd be glad to add this feature to `faker`.

Good suggestion, thanks!

ClarenceL:
This needs to be added, for one project we wanted to generate sandbox Stripe accounts for our test server but Stripe even in sandbox mode requires valid zip codes. Generating a set of address fields that are valid at least by city, state, zip would be really great.

karllhughes:
I am going to start working on this. My current thought is to use [davglass/zipcodes](https://github.com/davglass/zipcodes) to support something like this:

```
// Lookup zip code by state
faker.address.zipCodeByState('CA');

// Lookup zip code by city and state
faker.address.zipCodeByCityAndState('Cupertino', 'CA');
```

Welcome to feedback on this strategy though!

karllhughes:
I ended up not using @davglass's repo as I didn't want to introduce a new dependency and it doesn't account for localization in the same way as we would need. The PR I just made (#600) implements a `zipCodeByState` method that uses a new descriptor file to find min and max zip codes based on the two-letter state abbreviation.

mkiddTempus:
Some of the min max numbers are incorrect as they left off the leading `0`. For example: CT, MA, ME, NH,  NJ, NY, RI, and VT.

I looked at the tests and none of the selected states zip codes would have this issue.

owens-ben:
This is still an issue. I'm not even sure what the point of zipCodeByState() is, it doesn't seem to work even a little bit.

PR #600, opened 2018-01-20
Zipcode by state method

karllhughes:
@willyraedy and I worked on this to address #400.

We added a new method called `zipCodeByState` that takes a two-letter state abbreviation then uses the `postcode_by_state` definitions file (which was in the repo but not being used) to look up max and min values for zip codes in that state based on [the table @Marak pointed to](https://github.com/Marak/faker.js/issues/400#issuecomment-278768376).

I wasn't sure if we should commit the built files, but it seems that others didn't include them in their PR's so I figured that was part of the automated build process. Let me know if that's not right and I'll be happy to run a build and push updates.

Thanks for the great open source resource, hope this helps!


Marak:
@karllhughes @willyraedy 

Merged into `master`

This is a good addition, thank you for your contributions.

mkiddTempus:
There was a bug in this commit.

Some of the min max numbers are incorrect as they left off the leading 0. For example: CT, MA, ME, NH, NJ, NY, RI, and VT.

I looked at the tests and none of the selected states zip codes would have this issue as it only affects east coast areas that were the first to receive zip codes.

ATMartin:
We ran into an issue using this method recently: for anyone else who's confused about why they're getting invalid zip codes, **make sure your `faker.locale` is set to `en_US`.** Faker.js's default is `en`, which doesn't include US state zip code support. 

This method is currently made to support a single locale. Adding a warning/error if a valid-but-unsupported locale is used might be a nice addition - I'll keep a PR in mind as time permits! :)

PR #609, opened 2018-02-24
Firstname and lastname are not read properly for locale 'SV'

pvsenan:
When locale is set to sv, firstname and last name is not read from locale file. Fixed the issue and added more Swedish names with special Swedish characters.


pvsenan:
@Marak I don't remember exactly. But I can check it again. If it is not need I'll delete the old two files.

Marak:
Thank you @pvsenan 

Sorry for the delay in reviewing this.

We appreciate your help !

pvsenan:
@Marak I've removed the files that are no longer used.

PR #618, opened 2018-03-18
Added hungarian locale

sarkiroka:


Marak:
@sarkiroka 

I'd like to merge this PR, but I cannot merge in it's current state.

A couple things to start:

1. This PR adds 25,000+ new lines of code for a single locale. This is too large. Can we remove the `lorem` file completely? I see that first and last names are also several thousand lines. Can we reduce first and last name to smaller list?

2. I see we have several changes to core library code, but no updated unit tests. Can we add tests that will cover these newly added code paths?

Please let us know. I would like to see `hu` locale added.

sarkiroka:
@Marak, I reduced the names files and the lorem file. I think now manageable size of code.

The core modification need to use hungarian format of names and addresses, and remove some hardcoded strings. For add new tests, I agree it is needed, but I cannot help. I don't understand every logic, and it's too much time for me. So I would be glad if you complete the missing test

lib/locales/hu/address/state.js
```diff
@@ -0,0 +1,22 @@
...
+  "Nógrád megye",
+  "Heves megye",
+  "Jász-Nagykun-Szolnok megye",
+  "Csongrád megye",
+  "Békés megye",
+  "Hajdú-Bihar megye",
+  "Borsod-Abaúj-Zemplén megye",
+  "Szabolcs-Szatmár-Bereg megye",
+  "Budapest"
+];
```
bgolyoo, line 22: @sarkiroka missing `Somogy megye` from the list

lib/locales/hu/address/county.js
```diff
@@ -0,0 +1,21 @@
...
+    "Nógrád megye",
+    "Heves megye",
+    "Jász-Nagykun-Szolnok megye",
+    "Csongrád megye",
+    "Békés megye",
+    "Hajdú-Bihar megye",
+    "Borsod-Abaúj-Zemplén megye",
+    "Szabolcs-Szatmár-Bereg megye",
+    "Budapest"
+];
```
bgolyoo, line 21: @sarkiroka missing `Somogy megye` from the list

PR #636, opened 2018-04-30
removed name

SelaO:


Marak:
Again, you must not modify build or example files.

You must only modify source files.

Probably only going to be a few files in the end.

Closing again.

SelaO:
I asked you last time what folders to exclude, I don't appreciate this game you play with me. 

You could've saved us both time with this simple find and replace but I guess you prefer to immortalize the name of a Nazi war criminal. You should add Adolf Hitler too.

Marak:
I'm editing your comment and blocking you from contributing to this repository.

This is not a place to discuss politics or world history and I am not obligated to teach you how to make modifications to this project. 

This is not a game. If you don't understand how to modify the source code or behave civilly over Github Issues you may take your contributions elsewhere.

thewhodidthis:
Dear @SelaO, faker.js is a JavaScript port of [faker](https://github.com/stympy/faker) the Ruby gem, which in turn is based on Perl's Data::Faker. Please take a look at the [original list of names](https://github.com/stympy/faker/blob/master/lib/locales/en/name.yml) used for seeding. That might explain how Eichmann ended up being included over here as well and why your comments came across as confrontational even though you have a point, x

PR #648, opened 2018-05-22
Handle negative maximum with no minimum (random.number)

### LINKED ISSUE #647: faker.random.number() generates incorrect result when provided no minumum and negative maximum
When provided a negative maximum and no minimum, faker.random.number() will reliably produce a value inclusively in the range -10,0.  This is due to the library defaulting to 0 when min is not specified.

To reproduce:
```node -e "console.log(require('faker').random.number({ max: -10 }))"```

Pull request is forthcoming.

### PR DESCRIPTION
**Pull request by bjfelton**
Fixes #647.  Given a negative maximum and no minimum value,
return a value that is consistently <= max.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -30,13 +30,13 @@ function Random (faker, seed) {
 
     options = options || {};
 
-    if (typeof options.min === "undefined") {
-      options.min = 0;
-    }
-
     if (typeof options.max === "undefined") {
       options.max = 99999;
     }
+    if (typeof options.min === "undefined") {
+      options.min = options.max < 0 ? -99999 : 0;
+    }
+
     if (typeof options.precision === "undefined") {
       options.precision = 1;
     }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Marak** [comment]
Good catch! Will be merged for next release.

File: lib/random.js [hunk 1]
```diff
@@ -30,13 +30,13 @@ function Random (faker, seed) {
...
 
-    if (typeof options.min === "undefined") {
-      options.min = 0;
-    }
-
     if (typeof options.max === "undefined") {
       options.max = 99999;
     }
+    if (typeof options.min === "undefined") {
+      options.min = options.max < 0 ? -99999 : 0;
```
Marak, line 37: Sorry to nitpick @bjfelton, would you mind updating this so it doesn't use a conditional assignment operator? We try to avoid those in `faker.js`.

If you can make this small change I can merge this in.

PR #649, opened 2018-05-23
Feature/646 currency based on context

tobi-or-not-tobi:
Fixes #646. Provides both context to the currencyName and currencySymbol as well as a new currency method that returns a random object. I personally would only need the latter.

### DIFF
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -132,14 +132,26 @@ var Finance = function (faker) {
       return faker.random.objectElement(faker.definitions.finance.currency)['code'];
   };
 
-  /**
-   * currencyName
-   *
-   * @method faker.finance.currencyName
-   */
-  self.currencyName = function () {
-      return faker.random.objectElement(faker.definitions.finance.currency, 'key');
-  };
+    /**
+     * currencyName
+     *
+     * @method faker.finance.currencyName
+     * @param {string} [code] - a currency code
+     */
+    self.currencyName = function (currencyCode) {
+        if (currencyCode) {
+        var currencyName;
+            for(var prop in faker.definitions.finance.currency) {
+                if (currencyCode === faker.definitions.finance.currency[prop]['code']) {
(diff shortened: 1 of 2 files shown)


tobi-or-not-tobi:
I've not committed the generated build files (~100) as I wasn't sure if that was the intention.

tobi-or-not-tobi:
btw, during testing of the examples/browser/index.html I noticed an error with the creation of the API buttons. I've fixed that locally, but didn't want to pollute this PR.

tobi-or-not-tobi:
@Marak can you enlighten me on the code formatting for this project? I wasn't able to set it up in vscode. I've installed jshint fir this project (given the .jshintrc), but that showed a large number of errors. What are you using yourself? Perhaps something you can add to the contribution guide?

lib/finance.js
```diff
@@ -124,33 +124,77 @@ var Finance = function (faker) {
...
-          symbol = faker.random.objectElement(faker.definitions.finance.currency)['symbol'];
+      if (currencyCode) {
+        for(var prop in faker.definitions.finance.currency) {
+            if (currencyCode === faker.definitions.finance.currency[prop]['code']) {
+                symbol = faker.definitions.finance.currency[prop]['symbol'];
+                break;
+            } 
+        }
+      } else {
+          while (!symbol) {
```
Marak: Same issue here.

I'd prefer to use be using `while` or `break` keywords if necessary.

I understand there may or may not be a performance difference here in regards to speed, but I'd prefer clearer code over potential performance ( especially if we don't have benchmark data proving which is going to be faster ).

Marak:
@tobi-or-not-tobi -

I've added some comments for review.

In regards to linting rules, we may have fallen a bit behind. I will review the codebase myself on the next release to try and fix this.

For now, I would use the `.jshint` we have and try to just run it on the files you've modified.

Issue #663, opened 2018-07-01
DE - streetName(), streetAddress(), streetSuffix() return street names from EN locale


Radiergummi:
I've tested a few other german methods (`faker.address.country`, `faker.address.state`, `faker.address.firstName`, `faker.address.lastName`, ...) which all seem to work properly, it's only the street address methods that seem to fallback to the english locale. As there [hasn't been a commit to these files](./lib/locales/de/address) in the last three years, I suspect something else is going wrong here...

Radiergummi:
This can even be checked using the `faker.hook.io` instance:  
`http://faker.hook.io/?property=address.streetName&locale=de`

As you can see, it returns street addresses form the en locale.

Marak:
If you have tests which you've run that are failing please add the failing tests to the project in a PR and we can see about fixing the API if it's not working as intended.

Radiergummi:
Writing tests for this case would mean writing tests for all locale implementations which I guess isn't what you want. The way I quickly verified this was by doing `curl http://faker.hook.io/?property=address.streetName&locale=de` and grepping the second word in the english city suffixes, which matches every time. This, and the street format is `<num> <street>` instead of `<street> <num>`.

Marak:
We'll need a failing test in order to fix what you are reporting.

Please submit a PR including reproducible code of the problem or a potential code fix for the problem.

Thanks!

Radiergummi:
I've submitted the PR that highlights the problem in a test as you requested, even though I've got no clue on fixing it without digging further into the code.

Marak:
Localities in `faker` will fallback to english if the selected locality doesn't have locale data for that specific language.

Have you tried checking the locales folder to see what might be missing for German? It's possible we'll need to add the missing data, any help would be appreciated. 

I know a bit of German, so I should be able to review the changes. Auch, Das Leben ist kein Ponyhof.

Marak:
Related #133 #645 #337 #232 

Related: https://github.com/Marak/faker.js/blob/master/lib/address.js#L109

I think the issue is that the `address` module isn't using any of the custom formatting strings we previously didn't have available. It's the same issue across a few other API methods and localities.

It might just be a matter of making sure `address` can use data like https://github.com/Marak/faker.js/blob/master/lib/locales/de/address/street_address.js in formatting address strings.

Radiergummi:
I'm currently working on some other parts of our application but I'll try to check as soon as possible... Thank you for the investigation!

[...]

PR #664, opened 2018-07-01
Added test for broken german street address format

Radiergummi:
This adds the test required in #663

### DIFF
--- a/test/address.unit.js
+++ b/test/address.unit.js
@@ -159,6 +159,13 @@ describe("address.js", function () {
             faker.random.number.restore();
         });
 
+        it("formats using the correct locale format", function() {
+          faker.locale = 'de';
+
+          var street_address = faker.address.streetAddress();
+          assert.ok(street_address.match(/^\d+/) === null);
+        });
+
         context("when useFulladdress is true", function () {
             it("adds a secondary address to the result", function () {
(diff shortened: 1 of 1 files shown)


test/address.unit.js
```diff
@@ -159,6 +159,13 @@ describe("address.js", function () {
             faker.random.number.restore();
         });
 
+        it("formats using the correct locale format", function() {
+          faker.locale = 'de';
+
+          var street_address = faker.address.streetAddress();
+          assert.ok(street_address.match(/^\d+/) === null);
```
Marak: What does this regex check for?

Radiergummi: Whether the address line starts with one or more digits (US format, "1234 example st."), contrary to the expected non-digit characters (german format, "examplestr. 1234").

Marak: I'm not sure this is going to be the best way to test for the issue but it is helpful in showing one of the symptoms of locality fallback for locales with missing data.

Radiergummi: I think this specific test is highly convoluted and (crudely so) written for this exact problem, so I think I'd rather try to come up with something more generic, that you could use to determine how much of faker.js' API is supported for a language (like, "German: 98%")...

PR #676, opened 2018-08-03
Add times methods to Random module

mupkoo:
I find it useful to have the ability to run a given logic random times. For example, if you want to generate some data:

```javascript
faker.random.times(() => User.create());

faker.random.times(10, () => {
  // ... logic
  // ... logic
});
```

Not sure about the signature of the method though. Shall I use the options as a second argument?


Marak:
Thank you for your contribution! This PR is well structured.

I'm not sure about the actual API method which is being added. It seems like `random.times` isn't actually returning any fake data? Rather it is just executing a callback a random number of times?

Closing for now. I'm willing to re-evaluate this PR if anyone can justify the usage a bit better.

Marak:
Feel free to modify the PR and we can try it out. 

I'm open to the idea of having an array of results.

mupkoo:
@Marak I updated the PR. `faker.random.times` now will return the result from the given callbacks.

Take a look a let me know what do you think

test/finance.unit.js
```diff
@@ -162,7 +162,7 @@ describe('finance.js', function () {
         });
 
         it("should use the default decimal location when not passing arguments", function () {
-
+            faker.finance.amount().toString();
             var amount = faker.finance.amount().toString();
```
mupkoo, line 166: The first run returns `59.8`, that's why the second one

PR #683, opened 2018-08-15
Ensure faker.random.word does not return more than one word

### LINKED ISSUE #602: faker.random.words() returns phrases, not words
```
> faker.random.words(4)
```

### LINKED ISSUE #661: Random word (singular) sometimes returns more than one word

leviathanbadger:
faker.random.word() has a bug where sometimes it returns a string with more than one word. This PR fixes that bug by splitting the previous result on spaces and selecting randomly from the words returned.

Fixes #661 
Fixes #602

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -180,8 +180,8 @@ function Random (faker, seed) {
 
     // randomly pick from the many faker methods that can generate words
     var randomWordMethod = faker.random.arrayElement(wordMethods);
-    return faker.fake('{{' + randomWordMethod + '}}');
-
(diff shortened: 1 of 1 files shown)


Marak:
This seems good @aboveyou00 

Would you be able to add a new unit test for this?

Thank you.

Marak:
It seems like many users have had this issue, but we still don't have a unit test?

#661 
#602 

Going to merge since it's seem like this should work and resolve those other issues.

Thank you for your contribution @leviathanbadger !

PR #690, opened 2018-09-23
Update last_name.js

driade:
Corrections. Bad orthography, some bad last names.


Marak:
Thank you for your contribution @driade

PR #694, opened 2018-09-26
Add mixed options shape to the documentation

Tomassito:
From the documentation it is not clear what shape should the options object take.


Marak:
Merged.

Thank you for your contribution @Tomassito

PR #696, opened 2018-09-27
fix: ammend tests for bitcoin addresses

ashpr:
The tests for bitcoin addresses had an invalid regex.

@Marak I've left a note here explaining why we use 24,33 and not 25,34 as I can see why this would be confusing to anyone not too familiar with Regex... but if you think this is unnecessary I can remove it.

The regex quantifier only checks the preceding token so we do not need to check the length including the first character which is already tested via [13]


Marak:
This seems like it's right, but I'm not 100% sure.

Best we merge it and see if the tests break again or anyone complains.

Thank you for looking into this @ashpr, it's much appreciated it.

PR #698, opened 2018-10-01
Update Vietnamese locale data

thanhpd:
- Updated cities list
- Added country names in Vietnamese
- Removed county as Vietnam does not have a county-like system
- Added correct postal code format (5 digits since 2018)
- Updated correct phone number & cell phone format (Oct 2018)
- Added date/month in Vietnamese


Marak:
@thanhpd -

Can you confirm the change to `lib/locales/vi/phone_number/formats.js`?

I see we've removed some data here, was that intentional? Was there a problem with the existing phone number formats that have been removed?

thanhpd:
@Marak 
You can follow this article in Wikipedia: https://en.wikipedia.org/wiki/Telephone_numbers_in_Vietnam
Wiki can be unreliable but it's updated for both landline number and cell phone number, you can follow the reference links as well.

As for the removed phone number formats in `lib/locales/vi/phone_number/formats.js`:
```
  "01#### #####", -- This is the old cell phone number format and was discontinued
  "01### ######", -- Same as above
  "01#1 ### ####", -- Same as above, the 1 digit in 4th position from the left is not in any rules
  "011# ### ####", -- Same as above, the 1 digit in 3rd position from the left is not in any rules
  "03## ### ####", -- The 03 prefix was removed in landline format, this is the new prefix for cell phone number but there are only 10 digits
  "055 #### ####", -- Old area code
  "056 #### ####", -- Old area code
  "0800 ### ####", -- There's no area code starts with 0800
  "08## ### ####", -- Old area code
  "09## ### ####", -- 09 prefix is for cell phone number
  "016977 ####", -- 01 prefix is for cell phone number
  "01### #####", -- 01 prefix is for cell phone number
  "0500 ######", -- Old area code
  "0800 ######" -- There's no area code starts with 0800
```

Marak:
Thanks again for your contributions @thanhpd 

This additional information helps.

Merged.

PR #703, opened 2018-10-06
corrected the format

iUtsavB:
The actual format of mobile phone no is rarely starts with less than 6
so corrected that


Marak:
What about non-mobile phones @iUtsavB? Many Indians still have landlines.

This seems to actually remove a large block of potential Indian numbers?

iUtsavB:
ohh then that would be whole lot of data @Marak 
landlines have city codes, and if you generate landline numbers, it is to to find the actual owner name
so my suggestion is ignore landline numbers by the way landlines are kinda extinct

Marak:
I lived in Indian for 5+ years and saw a landline at almost every home and business I visited.

I've even spent several days having to navigate through my local BSNL office.

If anyone else from the region would like to comment I'd be glad to reconsider merging this PR.

Leaving open for now.

iUtsavB:
okay 😆 as you wish, will make a separate list for that.

iUtsavB:
Hello, @Marak you wanted supporting data for 1% figure, I was wrong it is 1.92% see [TRAI Data ](https://www.trai.gov.in/sites/default/files/PRNo98Eng18092018_0.pdf)

Marak:
@iUtsavB - You seem to be right. Thanks for the additional data!

I'm now reading closer to 2% land lines, but that is still much smaller then I thought. Interesting! My gut still tells me it's a bit higher than 2%, but also if you consider dense city populations it makes sense there are many more mobiles.

Marak:
Merged.

PR #708, opened 2018-10-06
Remove unnecessary whitespace in script tags

SheetJSDev:


Marak:
Hello @SheetJSDev 

As per [our contributing guide](https://github.com/Marak/faker.js/blob/master/CONTRIBUTING.md):


> The documentation is auto-generated, based on build/src markdown sources. If you modify the main Readme.md file, the Pull Request will be rejected, since it will be overwritten by the upcoming gulp execution

SheetJSDev:
FWIW starting from the current HEAD, the current Readme.md cannot be reproduced using the build script.  To verify:

```bash
$ git clone https://github.com/Marak/faker.js/
$ cd faker.js
$ npm install
$ npm run-script build
$ git diff Readme.md
```

There are also some untracked files as well a bunch of modified files with nontrivial (as in, not just the build date) changes.  One locale file in particular:

```diff
diff --git a/locale/nl_BE.js b/locale/nl_BE.js
index 67b5737..5478ea0 100644
--- a/locale/nl_BE.js
+++ b/locale/nl_BE.js
@@ -1,5 +1,5 @@
 var Faker = require('../lib');
-var faker = new Faker({ locale: 'nl_BE', localeFallback: 'nl' });
+var faker = new Faker({ locale: 'nl_BE', localeFallback: 'en' });
 faker.locales['nl_BE'] = require('../lib/locales/nl_BE');
-faker.locales['nl'] = require('../lib/locales/nl');
+faker.locales['en'] = require('../lib/locales/en');
 module['exports'] = faker;
```

SheetJSDev:
Probably better to first align the build system before making changes

Marak:
@SheetJSDev 

You are correct. 

Most likely the build script is only run on version releases and we then check in the changed files. 

The best approach for development on the project is probably going to be not checking in or worrying about any of the files which `gulp` is going to generate from it's build process. 

This means `master` generated files will not usually be 100% in sync with current API. This should not be a show-stopping issue and is intended by design.

Our Contributing document should be more clear in explaining this and I will improve our contributing documentation to clarify the process.

I would suggest only pushing commits to the repository that change source files. When the new release candidate is ready I will run the build script as part of our QA cycle and generate the new source of truth for the upcoming version.

FYI - I think there have been a few commits to fix the `ReadMe.md` formatting since the last release and I will make sure we get your fixes in if they are not already.

Hope that makes sense! Looking forward to seeing further commits from you.

PR #709, opened 2018-10-07
build script support for localeFallback

SheetJSDev:
`lib/locales/*/index.js` export `localeFallback` to override the fallback.  Only `nl_BE` currently uses this feature, and that locale was changed appropriately.

### DIFF
--- a/build/gulpfile.js
+++ b/build/gulpfile.js
@@ -157,11 +157,12 @@ gulp.task('nodeLocalRequires', function (cb){
   for (var locale in locales) {
     var localeFile = path.normalize(__dirname + "/../locale/" + locale + ".js");
     var localeRequire = '';
+    var localeFallback = locales[locale].localeFallback || "en";
     localeRequire += "var Faker = require('../lib');\n";
-    localeRequire += "var faker = new Faker({ locale: '" + locale + "', localeFallback: 'en' });\n";
+    localeRequire += "var faker = new Faker({ locale: '" + locale + "', localeFallback: '" + localeFallback + "' });\n";
     // TODO: better fallback support
     localeRequire += "faker.locales['" + locale + "'] = require('../lib/locales/" + locale + "');\n";
-    localeRequire += "faker.locales['" + 'en' + "'] = require('../lib/locales/" + 'en' + "');\n";
+    localeRequire += "faker.locales['" + localeFallback + "'] = require('../lib/locales/" + localeFallback + "');\n";
     localeRequire += "module['exports'] = faker;\n";
     fs.writeFileSync(localeFile, localeRequire);
   }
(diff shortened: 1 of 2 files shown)


Marak:
Very nice @SheetJSDev

I completely missed this! Blame https://github.com/Marak/faker.js/commit/e25e1ebc349c426c6b2598dc907eae497ef67160

This will be merged after I test it locally. 

Please let us know if you can expand on this commit or figure out a way to get this code path tested with our test suite.

Also, feel free to let us know if there are any other issues or questions in our API you'd like to address.

SheetJSDev:
@Marak to be clear, this is not an implementation issue in the library itself.  Problem lies in the build infrastructure.  [The `gulpfile` has a TODO](https://github.com/Marak/faker.js/commit/e25e1ebc349c426c6b2598dc907eae497ef67160#diff-b20c7230de4da834e815c06534cab810R136).

The first locale to use a non-`"en"` fallback was `nl_BE`, added much later in https://github.com/Marak/faker.js/commit/afcf8657dc996852a669723ac042382dc5986c61 .

Marak:
@SheetJSDev 

Thank you. I think I understand.
 
I also think we have no way to create a failing test for this since we have no [smoke tests](http://softwaretestingfundamentals.com/smoke-testing/) for generated build scripts?

Essentially, it would be nice if we could add a failing test that would become passing after merging this PR.

SheetJSDev:
If you really want a "test", run the build script again and compare the old and new files, essentially `npm run-script build && git diff --quiet lib`.  This doesn't work for some of the documentation (which have explicit timestamps) but none of the generated source files have a timestamp.

Marak:
@SheetJSDev 

I'm sorry, I must not have explained myself well or I'm confused.

I'd like a unit test which is going to test the ability to specify a new default locale besides `en`, and also this same test for the generated files ( to ensure no bugs in `gulp` script ).

I'm not sure I really care about the results of running a `git diff` here.

SheetJSDev:
Ok those are different issues :)

The "bug" that this fixes involves reproducing the generated files.  The proposed check (with `git diff`) ensures the builds are reproducible.

You are asking about testing the locale fallback.  That requires a parent locale PARENT, a child locale CHILD that falls back on PARENT, and a definition DEF that PARENT defines but CHILD does not.  In the current locale set, there is exactly one case: (PARENT=`nl`, CHILD=`nl_BE`, DEF=`lorem.words`).  Unfortunately, the `nl` version of `lorem.words` is identical to `en` version, so there is currently no way to test.

Marak:
It would be great if anyone could figure out a way to resolve this issue of adding testing for locale fallback.

I'm guessing we'll need more data points and perhaps some code updates.

Marak:
Closing this for now.

We now have a new build process and will be updating the structure of the library in order to better accommodate locales and fallbacks.

PR #711, opened 2018-10-09
fix(date): Allow for `refDate` in `soon()`, `recent()`

### LINKED ISSUE #686: Certain faker date functions do not accept a refDate
https://github.com/Marak/faker.js/blob/3a4bb358614c1e1f5d73f4df45c13a1a7aa013d7/lib/date.js#L71-L72

This was the one that caused me to take a look at the source, as it was behaving differently than `date.past`. As it stands now, this binds all of my seeded fake values to the current date and time. I am able to pass a `dateRef` to other functions such as `Date.past`, and I would like to do the same here.

There are two other functions that do not accept a `dateRef` as well, and therefore will change value depending on what time you start the application that is using fake data:

 - `Date.between`
 - `Date.soon`

Droogans:
Closes #686

There might be more date functions that need updating, if so just let me know what I missed.

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -12,7 +12,11 @@ var _Date = function (faker) {
    * @param {date} refDate
    */
   self.past = function (years, refDate) {
-      var date = (refDate) ? new Date(Date.parse(refDate)) : new Date();
+      var date = new Date();
+      if (refDate !== undefined) {
+          date = new Date(Date.parse(refDate));
+      }
+
       var range = {
         min: 1000,
         max: (years || 1) * 365 * 24 * 3600 * 1000
@@ -33,7 +37,11 @@ var _Date = function (faker) {
    * @param {date} refDate
    */
   self.future = function (years, refDate) {
-      var date = (refDate) ? new Date(Date.parse(refDate)) : new Date();
+      var date = new Date();
+      if (refDate !== undefined) {
+          date = new Date(Date.parse(refDate));
+      }
+
       var range = {
         min: 1000,
         max: (years || 1) * 365 * 24 * 3600 * 1000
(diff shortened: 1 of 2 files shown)


lib/date.js
```diff
@@ -67,9 +67,10 @@ var _Date = function (faker) {
    *
    * @method faker.date.recent
    * @param {number} days
+   * @param {date} refDate
    */
-  self.recent = function (days) {
-      var date = new Date();
+  self.recent = function (days, refDate) {
+      var date = (refDate) ? new Date(Date.parse(refDate)) : new Date();
```
Marak: Sorry to nitpick, would you mind updating this PR to not use the `?` conditional operator.

I'd prefer to use `if` `else` and add a couple extra lines of code instead of using the `?` operator.

I've always personally found `?` statements hard to read and have been trying to not introduce them into `faker.js`

Marak:
@Droogans -

This looks good, thank you!

If there are additional date functions which need updating I would very much appreciate it if you could please look into it and try to fix them now ( since you are most recently familiar with these code paths today )

I've made one small comment about a code style update to not use the `?` operator. This isn't a show-stopper for getting the PR merged. I'd prefer to be able click "Merge Pull Request" on Github, but I can also locally make the change to help get this moving forward.

Thanks again for your contributions and feedback!

lib/date.js
Droogans: I'll pull them out of the rest of the lines in this file while I'm at it, too.

Marak:
Merged.

I ended up slightly changing the `undefined` check. JavaScript can be somewhat unwieldy and I've found it's best to use the `if ( typeof myVar === 'undefined' )` syntax or some odd edge cases may sneak in.

Please let us know if you find any other improvements for date / time functions. I believe there still might be some bugs from the previous implementation or room for further improvements.

Thank you @Droogans !!

PR #713, opened 2018-10-11
locale ru: add name.middleName with russian gender for russian passport (отчество)

jonny64:


Marak:
@jonny64 -

Would it be possible to add new test coverage for this code path? 

I don't speak Russian very well and I'm not sure we have many other languages that allow for gendered middle names. It will be difficult for me to verify these results.

__________

Спасибо за ваш вклад.

Не могли бы вы добавить больше тестового покрытия? Я плохо знаю русский. Я не могу проверить себя.

Я машина.

Marak:
We never got a confirmation of the translation, but I think @jonny64 did get it right the first try.

If anyone can find an issue with this PR, please let us know.

Thank you @jonny64 !!!

paulcervov:
@Marak and @jonny64 thanks for your work, I really missed these methods "out of the box" 👏
I checked the translation, everything is fine 👍

PR #714, opened 2018-10-14
Add locale pt_PT

titonobre:
Changes:
- Add locale pt_PT
- Add `male_prefix` and `female_prefix` to definitions whitelist


Marak:
@titonobre 

Thank you for your contribution. A couple of questions before we can merge.

Can you elaborate on the male / female suffixes? I've never seen this before for other languages and am not familiar. I ask because this will require adding new API property to main definitions ( which is OK ), but I'd like to understand why before adding.

Also, I'm seeing strange characters on my machine for the gendered suffixes. I've attached a screenshot. Are you seeing the same thing or is there something I need to do on my computer to get these to render correctly?

titonobre:
Hi @Marak.

In portuguese (and several other languages) we use different prefixes depending on whether the grammatical gender is masculine or feminine. Pretty much like Mister (Mr.) and Mistress (Mrs.).

- Senhor (Sr.) / Senhora (Sr.ª)
- Doutor (Dr.) / Doutora (Dr.ª)
- Engenheiro (Eng.) / Engenheira (Eng.ª)

Regarding those strange characters, they are perfectly rendered. 🙂 They are called [Ordinal Indicators](https://en.wikipedia.org/wiki/Ordinal_indicator) and are used, among other things, in abbreviations. We even have a dedicated key for them on our keyboards.

image
see [Portuguese keyboard layout](https://en.wikipedia.org/wiki/Portuguese_keyboard_layout)

The different symbols are used depending on whether the grammatical gender is masculine (º) or feminine (ª).

Despite all this, I'm now considering a different form of abbreviation. To collect the information for this answer, I read a few things about this subject and it seems that since 2009 the suggested way to abbreviate _Senhora_ is _Sra._ and not _Sr.ª_. Today I learned something new 😊. New commit on the way.

Marak:
> > I ask because this will require adding new API property to main definitions
> 
> Regarding the new properties, they are already being used here
> 
> [faker.js/lib/name.js](https://github.com/Marak/faker.js/blob/c852cdf5ba5253617fffc0261fa6c62e77a200a4/lib/name.js#L124)
> 
> Line 124 in [c852cdf](/Marak/faker.js/commit/c852cdf5ba5253617fffc0261fa6c62e77a200a4)
> 
>  if (typeof faker.definitions.name.male_prefix !== "undefined" && typeof faker.definitions.name.female_prefix !== "undefined") {

@titonobre -

Interesting! At a quick glance I would wonder if this is even working in `master`? I would assume your changes fix an existing problem? Could you possibly investigate to see if we can add some test coverage here? Regardless, I appreciate the additional information. Thank you.

titonobre:
I've added a few unit tests for `name.findName()` and `name.prefix()`.

Marak:
THANK YOU @titonobre 

This is very nice and good work. I can see prior to this PR our `lib/name.js` file is reporting `79.31%` code coverage....which is actually the worst file in the library!!!

I will pull down this PR locally and begin testing soon to see how the new tests work. Let me know about any additional changes based on your new research for prefix. I'm guessing this will be small change.

We really appreciate your contributions. Thanks again!

Marak:
Merged. 

Closing PR for now.

Please feel free to make any additional changes in a new PR.

PR #717, opened 2018-10-22
fix(npm): reduce package size

SimonSchick:
Relatively trivial.
You were shipping some redundant files (locale/.publish) and a few config files,
this saves roughly 4mb (disk size) not sure about package size because I can't get it to build correctly locally.

I renamed the license so npm picks it up automatically.

Please review that this includes all files you need/want, I'm sure I got all but you never know 🙂 

changelog, license, readme are automatically included


SimonSchick:
Where is travis? :c

Marak:
Thank you for your contribution @SimonSchick 

I think this has been resolved with a few other PRs.

Please feel free to check the latest release and see if we need to make any additional changes the ignore files.

Thanks again!

PR #718, opened 2018-10-23
Fix random module context when using a seed

Cliffzz:
Bind the methods for the random module to the correct context when using a seed.

For example this fixes

```javascript
const faker = require('faker');
faker.seed(1);
faker.fake('{{random.uuid}}'); //TypeError: self.number is not a function
```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -17,6 +17,15 @@
 
 */
 
+function bindAll(obj) {
+    Object.keys(obj).forEach(function(meth) {
+        if (typeof obj[meth] === 'function') {
+            obj[meth] = obj[meth].bind(obj);
+        }
+    });
+    return obj;
+}
+
 /**
  *
  * @namespace faker
@@ -37,15 +46,6 @@ function Faker (opts) {
   self.localeFallback = localeFallback;
 
   self.definitions = {};
-
-  function bindAll(obj) {
-      Object.keys(obj).forEach(function(meth) {
-          if (typeof obj[meth] === 'function') {
-              obj[meth] = obj[meth].bind(obj);
-          }
-      });
-      return obj;
-  }
 
   var Fake = require('./fake');
   self.fake = new Fake(self).fake;
@@ -156,6 +156,6 @@ Faker.prototype.setLocale = function (locale) {
 Faker.prototype.seed = function(value) {
   var Random = require('./random');
   this.seedValue = value;
-  this.random = new Random(this, this.seedValue);
+  this.random = bindAll(new Random(this, this.seedValue));
 }
 module['exports'] = Faker;


Marak:
Thank you @Cliffzz 

Is there a reason the `bindAll` method has been moved? Is this cosmetic? 

It would preferred to only merge the actual code changes.

Cliffzz:
`bindAll` was part of the `Faker` scope so to me it made the most sense to move it so it can be reused in `Faker` and in `Faker.prototype.seed`.

If you prefer another solution let me know.

Marak:
I don't think moving the function is going to make any actual difference here though? I'd prefer to merge the PR without the change since it will be confusing when reviewing history.

Either way this is a good catch and I will make sure to apply the fix for `faker.fake` before the next release.

Thank you.

Cliffzz:
> I don't think moving the function is going to make any actual difference here though?

Well without moving it `bindAll` is undeclared in `Faker.prototype.seed = function(value) {`

Alternatively if you prefer, I could move  `Faker.prototype.seed = function(value) {` into `Faker` so it will be like

```javascript
  self.seed = function(value) {
    var Random = require('./random');
    this.seedValue = value;
    this.random = bindAll(new Random(this, this.seedValue));
  };
```

Marak:
So I've just now tested this change locally and it appears it's not actually fixing the problem initially described?

This also seems like it could be a regression as well #368 #377 

I do believe there is an issue, but I am a bit confused here. 

Where else in the code are we using `self` reference inside modules? Why not just change https://github.com/Marak/faker.js/blob/master/lib/random.js#L146 to use `faker.random` scope?

Going to need more information on this or perhaps a failing test. Do you know anywhere else in the code-base where this scope issue is happening?

Thank you for the additional information. I do want to make sure this is fixed correctly.

Cliffzz:
I just had an issue with `random.uuid` and after taking another look it indeed looks like `self` is only used for `random.uuid` so using `faker.random` should suffice.

The reason I went for the `bindAll` route is because it looked like it was simply forgotten for `seed` in https://github.com/Marak/faker.js/pull/377

Marak:
@cliffpyles -

Thanks again for your contribution and for pointing out this issue with `random.uuid`.

I think the issue has now been resolved with https://github.com/Marak/faker.js/commit/ed9c560599bf1c32ddc36c2f71307c36b810c7c9

I'm doing a general review on the API now to see if we can remove `bindAll` entirely. I'm also adding more unit tests which covered this failing case.

Please let us know if you experience any other others.

cliffpyles:
@Marak I think you meant @Cliffzz

Marak:
I'm sorry @cliffpyles! I did mean @Cliffzz 

It seems I was too fast on Github autocomplete.

My apologies! 

:)

PR #721, opened 2018-10-27
Add 92 range number of MEO

simaolemos:


Marak:
Thank you for our contribution @simaolemos 

I don't fully understand the nature of this data change, but it seems simple enough to merge.

PR #728, opened 2018-11-14
Add gendered french first names

Kignuf:
This commit has two goals:
* Have more french first names so that I don't often get the same ones
* Allow to generate a male or female first name

This solves those two problems by:
* adding french first names found on wikipedia
* saving them separatly by gender (based on how 'ru' locale does it)

How to use:
```
faker.locale = "fr"
faker.name.findName(null, null, 1)
faker.name.firstName(0)
```


Marak:
Thank you for your contribution @Kignuf !!!

I will assume @deraw thumbs up is confirmation of translation accuracy.

Merged.

PR #731, opened 2018-11-23
Update currency symbol of turkish lira

scokmen:
Currency symbol of turkish lira updated

https://en.wikipedia.org/wiki/Turkish_lira_sign


Marak:
Thank you for your contribution @scokmen 

Merged.

PR #738, opened 2018-12-27
Change fallback for fr-CH to fr

### LINKED ISSUE #737: Better locale fallback for fr_CH
@jfreyre I see you added the `fr_CH` locale here, should we use `fr` as fallback instead of `en`? 

https://github.com/Marak/faker.js/blob/d3ce6f1a2a9359574e7f31f14d4901648047c45a/locale/fr_CH.js#L2

Floriferous:
Closes #737 

I'm not sure if these are the only changes that need to be made.


Marak:
@Floriferous -

You have the correct idea here, but I don't think this solution will work 100% correct with current `faker.js` structure, so I will have to close the PR. 

We are tracking this problem and should have a solution that works for all locales in the next major release version

Thanks again!

RE: #977

PR #739, opened 2018-12-28
Fixed faker.random.word() returning multiple words

### LINKED ISSUE #602: faker.random.words() returns phrases, not words

### LINKED ISSUE #661: Random word (singular) sometimes returns more than one word

Acanguven:
I know lots of people waiting for this.

**Problem**
`faker.random.word()` returns multiple words instead of returning a single word.

**Fix**
This PR fixes the problem by splitting the result from faker method and returns one of the array elements.

**Issues**
Fixes #661
Fixes #602

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -78,7 +78,7 @@ function Random (faker, seed) {
       }
       return faker.random.number(opts);
   }
-  
+
   /**
    * takes an array and returns a random element of the array
    *
@@ -202,7 +202,9 @@ function Random (faker, seed) {
 
     // randomly pick from the many faker methods that can generate words
     var randomWordMethod = faker.random.arrayElement(wordMethods);
-    return faker.fake('{{' + randomWordMethod + '}}');
+    var result = faker.fake('{{' + randomWordMethod + '}}');
+
+    return faker.random.arrayElement(result.split(' '));
 
   }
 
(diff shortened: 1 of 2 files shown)


Acanguven:
@marak Any updates?

yenereren:
any updates ? @Marak

erdoganoksuz:
Last week i spent 2 hour to understand why my tests fail.
Then i discern faker.random.word() sometimes returns multiple word.

@Marak If method name **word** instead of **words**, should not returns multiple words.

WoLfulus:
I'm waiting for this too

Acanguven:
At least add some people as maintainer so we can merge useful pull requests like this when you are away.

opravil-jan:
Will be this issue solved soon? It is every paintfull.

lib/random.js
```diff
@@ -202,7 +202,9 @@ function Random (faker, seed) {
 
     // randomly pick from the many faker methods that can generate words
     var randomWordMethod = faker.random.arrayElement(wordMethods);
-    return faker.fake('{{' + randomWordMethod + '}}');
+    var result = faker.fake('{{' + randomWordMethod + '}}');
+
+    return faker.random.arrayElement(result.split(' '));
```
Khartir: This alone won't be enough. I use this code in a project and it just returned `Islands,` (I think). So at least commas have to be trimmed of at the end, I don't know what else.

Marak:
Thank you for your contributions everyone.

I did a quick look and I don't think this will PR will work as expected.

We've got a few other solutions pending for this, so it should be resolved soon.

Closing and locking.

PR #748, opened 2019-01-23
add Japanese lorem data

tomoyamachi:
- add Japanese lorem data
- this data from https://github.com/stympy/faker/blob/master/lib/locales/ja.yml


Marak:
Thank you for your contribution @tomoyamachi !!!

We appreciate you citing the source of this data.

Merged.

PR #775, opened 2019-04-18
Get x dates between two dates

borel:
Hey Marak , thanks for this usefull package.

I suggest you this enhancement , who permit to get an array of date between two dates.

It could be usefull as core function or to be use to get an array of transaction for example.

Have a good day.


Marak:
@borel -

I can see we already have a `Date.between()` function in faker. How is this new function different?

It seems like we have duplication here?

borel:
Hey @Marak , 

Date.between() return only one date between two dates.
Date.betweens() return x dates  ( x is defined by the third and new parameter ) between two dates.

have a good day !!

Marak:
Thank you for your contribution @borel 

Merged.

PR #776, opened 2019-04-18
Closes #548. Return default country if set by locale

### LINKED ISSUE #548: Errors with Fr address
Hi,

1/ Street prefix.
In the file street_prefix.js there is : "Allée, Voie", instead of "Allée", "Voie",

2/ Country returns other than France.
There is no country definition and default_country is not define in the object _definitions

Best regards

alekstr:


lib/address.js
```diff
@@ -188,7 +188,9 @@ function Address (faker) {
    * @method faker.address.country
    */
   this.country = function () {
-    return faker.random.arrayElement(faker.definitions.address.country);
+    return (faker.definitions.address.country && faker.definitions.address.country.length > 0) ?
```
Marak, line 191: This line is a bit too long and we generally try not use conditional assignment operators in `faker.js`

PR #777, opened 2019-04-21
Add 'fr' and 'ja' locale country names and adjust 'pt_BR' city suffixes

tsilvap:
Regarding the second commit: I added a space before each 'pt_BR' city suffix, in order to fix the 'pt_BR' city generation. Right now, address.city for 'pt_BR' returns names such as "Vila Heitordo Norte", when it should be "Vila Heitor do Norte" (the "do" should always be separated from the previous word).


Marak:
@tsilvap -

Would it be possible if you can resolve the conflicts on this PR so we can merge?

FYI, usually we try to only update one locality at a time. It can get a little confusing otherwise. Thanks again!

Marak:
Thank you for your contributions @tsilvap 

Merged!

PR #792, opened 2019-06-09
Fix #663

### LINKED ISSUE #663: DE - streetName(), streetAddress(), streetSuffix() return street names from EN locale

PeachIceTea:
This fixes #663. 
It also adds faker.address.buildingNumber. I personally already used it and since its used internally already I felt it wouldn't be an issue.
I also had to add a new helper faker.helpers.templateString. It deals with the #{} template strings used in the street_root.js files. I didn't find any other function that do this. I am however not familiar enough with the codebase to be sure.

### DIFF
--- a/test.js
+++ b/test.js
@@ -1,0 +1,31 @@
+#!/usr/bin/env node
+const faker = require(".");
+faker.locale = "de";
+
+console.log(faker.definitions.address.building_number)
+
+function telefonNummer() {
+  return "0".concat(
+    faker.phone
+      .phoneNumber()
+      .slice(4)
+      .split("-")
+      .join(" ")
+  );
+}
+
+const firstname = faker.name.firstName();
+const lastname = faker.name.lastName();
+const adress = faker.address.streetAddress().split(" ");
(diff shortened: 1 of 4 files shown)


test.js
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env node
```
Marak, line 1: @PeachIceTea - `faker.js` has a built-in test suite with `mocha`. All tests will need to be put there.

FabianTe:
@Marak Hi, I'd like to contribute to this fix since I need the german street names. As far as I understand this PR is fine except that the test.js file is not wanted, correct?

Marak:
Closing due to no tests and inactivity.

PR #793, opened 2019-06-10
Update options.max for random.number to Number.MAX_SAFE_INTEGER

frytyler:
Hey, my team has been using `faker.js` for a while and really love it. We had a discussion internally about `faker.random.number` and noticed the max was `99999` and talked about using it like this instead for safety. 

```javascript
faker.random.number({ max: Number.MAX_SAFE_INTEGER });
```

I suggested that we try and update the default here so we can avoid that potentially multiple times in out codebase. 

No strong opinion here if there was reasons for going with `99999` originally. 

**Note** I took a look at the unit tests and noticed there was no test for the default max value so I wasn't sure how to add a test for it here.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -35,7 +35,7 @@ function Random (faker, seed) {
     }
 
     if (typeof options.max === "undefined") {
-      options.max = 99999;
+      options.max = Number.MAX_SAFE_INTEGER;
     }
     if (typeof options.precision === "undefined") {
       options.precision = 1;


frytyler:
Interesting, maybe this won't work then. Node `0.10` may not support this.

Thoughts?

frytyler:
Closing this as I haven't heard anything from the maintainers so my assumption is no interest. No point in it just sitting stale

Marak:
I don't fully understand the implications of changing `99999` to  `Number.MAX_SAFE_INTEGER`, so I am hesitant to make this change.

If anyone can better explain what this change would actually do in `faker.js` ( including potential problems )  I am glad to review this again.

okka-riswana:
Hi, I am not sure if this has been discussed again in another place, but I have encountered an issue about this in our team. Trying to generate a random number with a minimum value above `99999` resulted in a value lower than expected.

For example, our project deals with a lot of big numbers above one million. So trying to generate a random number with the `min` option set to a bigger value with no `max` option always gave out values lower than `99999`.

```js
console.log(random.number({ min: 1000000 }))
// 196134
```

But, I probably could see why assigning `Number.MAX_SAFE_INTEGER` as the default option for `max` would lead to problems such as getting too big of a number if the `max` option is not set.

Also, `Number.MAX_SAFE_INTEGER` is quite a big number that the random values might overflows 32-bit integer type. So perhaps, we could choose a value much closer to the max 32-bit signed integers (`2147483647`) as `99999` has chewed through 16-bit ints already.

PR #795, opened 2019-06-12
Extended postcode.js file in "locales\fr\address" with support for only french zip codes

tamaravrzic:
Before it was defined only the format of the zip code, but it generated zip codes from all over the world. 
With this addition it only generates zip codes in France.
[postcode.zip](https://github.com/Marak/faker.js/files/3280739/postcode.zip)


Marak:
@tamaravrzic -

I'm not sure what happened here, but the PR is trying to change 696 files.

Closing for now. Please feel free to open a new PR with this code. Thank you!

Issue #699, opened 2018-10-01
Add support for faker.time()


siwalikm:
Hi @Marak, I couldn't find faker methods for faker.time() which is very useful at several use-cases.

If this seems to be something useful, I can raise a pull request for this.

Marak:
@siwalikm -

Sounds good

Will accept PR.

Please check `master` branch for any date or time related functions. I believe there are a few already.

siwalikm:
Alright Marak! I'm starting to work on it. Leaving this issue open so that anyone searching for this under issues will have some context that it is a WIP.

siwalikm:
Hi @Marak, I have an open PR for this issue for almost 6 months now. #729 
Wondering if you can merge it?

PR #807, opened 2019-07-12
Adding support for fake Time Method()

siwalikm:
@Marak With respect to Github issue #699 , raising this PR to add time method support for faker.

Right now this PR adds support for recent time.

```javascript
faker.time.recent(); 
// > 1562053376052

faker.time.recent('wide');
// > "13:13:28 GMT+0530 (India Standard Time)"

faker.time.recent('abbr');
// > "13:13:47"
```

Future scopes: `past/future time in range`, `random past/future time`

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -83,6 +83,9 @@ function Faker (opts) {
   var _Date = require('./date');
   self.date = new _Date(self);
 
+  var _Time = require('./time');
+  self.time = new _Time(self);
+
   var Commerce = require('./commerce');
   self.commerce = new Commerce(self);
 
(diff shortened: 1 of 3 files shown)


julientaq:
Hi!

This is what i need :) 
Is there a reasonm this merge request didnt make it to the code?
Thanks a lot!

Marak:
@siwalikm -

I had to update the test cases a little bit.

We were having some issues with the PR causing tests to randomly fail on the CI sometimes. The way the tests were setup left a race condition open for occasional test failures ( which is bad ). 

There is no guarantee the date objects will always be created at the same time. I would suggest in the future using something like `sinon` to stub out the date creation.

siwalikm:
Sorry for that @Marak. I'll use sinon to stub going forward.

PR #812, opened 2019-08-02
correct dobra code and symbol

gmunguia:
According to the following sources, STN is the ISO code of dobra:
* https://en.wikipedia.org/wiki/S%C3%A3o_Tom%C3%A9_and_Pr%C3%ADncipe_dobra
* https://www.xe.com/currency/stn-sao-tomean-dobra


Marak:
Good catch @gmunguia !

Thank you for your contribution. We appreciate you citing your source data.

Merged.

PR #816, opened 2019-08-20
Adding some names

valdirmendesgt:
- Adding some names


valdirmendesdev:
Hello @Marak, no, there is no specific reason.

The [state abreviation](https://github.com/Marak/faker.js/blob/master/lib/locales/pt_BR/address/state_abbr.js) file is actually missing some states. the other additions I made because I really wanted to contribute to the project.

But if you wish, I can close the PR. The project is already awesome.

Marak:
@valdirmendesdev -

We can merge this. Looking at the previous comment I think we were asking for additional data.

If you are able to resolve the conflicts I will hit the merge button.

Thanks again!

Issue #778, opened 2019-04-21
[Feature Request] add Video and Audio faker


ALiangLiang:
As title. I want a media faker to generate playable video and audio url.

Marak:
@ALiangLiang -

Thank you for your contribution. That sounds like a pretty good idea.

Can you elaborate a bit on the API you would like to see?

PR #828, opened 2019-09-20
Media url

edumoreira1506:
Resolves issue #778 

Well, I don't know any repository of audios, but for videos, we have the youtube, so I added one method than returns one random url video, this would be help some situations or specific projects. I hope that helps :)


Marak:
Tracking #918

Moosh-be:
hello @edumoreira1506  what's missing to finish this ?

Nikolay-Vovk-dataart:
I think its better to use some video that will be available for all countries. Ive tried to use the youtube samples - not allowed (Russia).

PR #837, opened 2019-10-09
Added the Music module

mrstebo:
This is a very simple faker module at the moment that only contains a `genre` method that will return a genre of music. This would be valuable for fake data relating to some kind of music site.

I will eventually add some more methods to this module so that it can become even more useful :+1:


Marak (approved):

Marak:
@mrstebo -

Merged!

Thanks again!

PR #838, opened 2019-10-09
faker.seed(0) should not be ignored

bwiklund:
`faker.seed(0)` is currently ignored, it probably shouldn't be?

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -6,7 +6,7 @@ var mersenne = require('../vendor/mersenne');
  */
 function Random (faker, seed) {
   // Use a user provided seed if it exists
-  if (seed) {
+  if (seed != null) {
     if (Array.isArray(seed) && seed.length) {
       mersenne.seed_array(seed);
     }


lib/random.js
```diff
@@ -6,7 +6,7 @@ var mersenne = require('../vendor/mersenne');
  */
 function Random (faker, seed) {
   // Use a user provided seed if it exists
-  if (seed) {
+  if (seed != null) {
```
Marak: That doesn't seem right.

bwiklund: ah, one sec, let me put this in the right place

bwiklund: wait no, this seems right, this is called straight from `Faker.prototype.seed`. this line is supposed to allow 0 to be used a seed, while throwing away null/undefined

bwiklund: @Marak does this make sense?

Marak:
Thank you for your contribution @bwiklund 

Closing for now since there are no additional tests.

I can see why this would be confusing for a user if `0` is ignored, but we should dig a bit deeper and make sure this will work as intended.

Please feel free to re-open with test.

PR #840, opened 2019-10-10
Fixed issue where `useAbbr` parameter not used.

mrstebo:
It will now return an abbreviated state if `useAbbr` is true.

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -207,7 +207,9 @@ function Address (faker) {
    * @param {Boolean} useAbbr
    */
   this.state = function (useAbbr) {
-      return faker.random.arrayElement(faker.definitions.address.state);
+      return useAbbr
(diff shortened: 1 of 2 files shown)


lib/address.js
```diff
@@ -207,7 +207,9 @@ function Address (faker) {
    * @param {Boolean} useAbbr
    */
   this.state = function (useAbbr) {
-      return faker.random.arrayElement(faker.definitions.address.state);
+      return useAbbr
+        ? faker.random.arrayElement(faker.definitions.address.state_abbr)
```
Marak, line 211: @mrstebo - 

Sorry to nitpick, can you remove the conditional assignment line here? We try to avoid those in the code base.

Thank you.

mrstebo: No problem. I have updated this :+1:

PR #841, opened 2019-10-10
Integrate ESLint and Husky

mrstebo:
I have noticed that there are some files that have a mix of standards. Using ESLint allows us to fix any issues when using `eslint --fix`.

**FYI: I have ran polyjuice to convert the existing .jshintrc to a .eslintrc config**

Now I know that having a manual task for contributors to run can be a bit of a pain, so I have also added the ability for it to be ran automatically using hooks.

Husky allows us to define some hooks. The hooks I have used are `pre-commit` and `pre-push`. 

The `pre-commit` hook runs `lint-staged`. This will ensure that the files that are committed follow the style guide. It will attempt to fix these automatically, and then stage the fixes.

Th `pre-push` command runs `npm test`. This ensures that whenever code is pushed the tests pass. This reduces the chance of build failures on the CI pipeline.

I've not ran `eslint --fix` on the entire codebase, as that would make this PR crazy big. But if this PR is accepted, then the option is there to start fixing some of these files :+1:


Marak:
I took a look through the code base and I want to get the linting fixed soon.

The first thing we have to do is get as many of the open pull requests merged first so we don't create a ton of conflicts by linting first.

Once we have a bit more of the PRs merged, I will spend a day fixing the linting.

PR #844, opened 2019-10-17
Add support for Nigerian locale

djunehor:
address
company
internet
phone_number
name


Marak:
Thank you for your contribution @djunehor 

This is great!!! I'm glad to add the new locale to `faker.js`

Merged. 

Would be good if we had anyone else familiar with this locale to review the data.

PR #847, opened 2019-11-02
fr/address/city_name.js: fix typo in "Lille" name

hydrargyrum:
"Lille" is the real city name, not "Lille13".


Marak:
Good catch @hydrargyrum 

Merged.

PR #853, opened 2019-12-06
Fixed the currency list

veritech:
Previously this list included invalid ISO4217 codes, probably as a result of the parser/generator that was used to create the list.

I've split all of the combined codes, and removed some of the [fund code](https://en.wikipedia.org/wiki/Unidad_de_Fomento) codes.

I've also taken the time to give some the currencies their correct names and symbols according to their respective wikipedia pages.


woeltjen:
Thank you for this! I started hitting `Invalid currency code: SVC USD` errors in unit tests, and finding this PR saved me from losing too much time trying to find the root cause. (It'll save more people more time once it gets merged 😁)

Marak:
Thank you for your contribution @veritech 

We had a slight merge conflict with another PR, but I believe yours was more accurate. Hopefully the merge went okay, I tried resolving this one directly on the Github website editor instead of my usual workflow.

Merged. Thanks again!

Issue #797, opened 2019-06-14
lorem image dead ?


crapthings:

darksh3ll:
yes  lorempixel dead 👍

crapthings:
which one should we use for replacement ?

unsplash ?

darksh3ll:
**I change url it works thank you**
node_modules => faker/lib/images
 var url = protocol + 'loremflickr.com/' + width + '/' + height;
      if (typeof category !== 'undefined') {
        url += '/' + category;
      }

bretanac93:
We can also integrate lorem Picsum which is another similar service, I can open a PR and make an integration the same way Lorem Pixel used to be, just give me green light :)

Unkopath:
Solved it changing these parameter on Faker/Provider/Image.php

the base url for:

`$baseUrl = "https://picsum.photos/"; `

and on the image function adding:

```
curl_setopt($ch, CURLOPT_FILE, $fp);
curl_setopt($ch, CURLOPT_SSL_VERIFYHOST, false); 
curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, false);
curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
curl_setopt($ch, CURLOPT_POSTREDIR, 3);
```

PR #855, opened 2019-12-15
Adds Lorem Picsum as image provider

abnersajr:
## Description

As suggested in #797 discussion, this PR adds [Lorem Picsum](https://picsum.photos/) as a faker Image Provider.
Lorem Picsum is reliable and stable, with DigitalOcean backing this project.
This can generate images in grayscale and blurred images too.
Another feature is to generate the same random image, based on a seed.

## Methods added

Under the namespace `lorempicsum`:

- image(width, height, grayscale, blur)
- imageGrayscaled(width, height, grayscale
- imageBlurred(width, height, blur)
- imageRandomSeeded(width, height, grayscale, blur, seed)

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -3,6 +3,7 @@
  * @namespace faker.image
  * @property {object} lorempixel - faker.image.lorempixel
  * @property {object} unsplash - faker.image.unsplash
+ * @property {object} unsplash - faker.image.lorempicsum
  * @default Default provider is unsplash image provider
  */
 var Image = function (faker) {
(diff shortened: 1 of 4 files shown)


Kageetai:
Can we get an update on this?
With LoremPixel still being down 99% of the time, this is a greats improvement. But from my understanding, lorempixel will still be used all of the time, no? Is there any way to change the provider when calling `image.imageUrl`?

Marak:
Thank you for your contribution @abnersajr 

This looks OK to me, but I have not fully tested the third-party service.

Merged.

Issue #652, opened 2018-06-05
use HTTPS for lorempixel.com


armadillojim:
Howdy!  I'm a long-time user, and first-time poster.  Thanks for the wonderful library!

I noticed today that the image URLs being generated for lorempixel.com use HTTP and not HTTPS.  It seems that they've supported SSL now for just over a year (since mid-March 2017).  See https://twitter.com/lorempixel/status/842398954767716352

Could you change all the `http`s to `https`s?  Or I can create a PR if that'd be welcome.

Thanks!

Marak:
Yes @armadillojim 

I'm just seeing this now. We should switch to HTTPS by default in the API.

I've asked LoremPixel if they would re-enable non-SSL connections so older versions of faker work again. https://twitter.com/marak/status/1044228671173775360

aexol:
By the way lorempixel is so slow :| Shouldn't we think of another image generation solution? 
@Marak

Marak:
@aexol -

Related issues: 

https://github.com/Marak/faker.js/issues/641
https://github.com/Marak/faker.js/issues/364

Potential PR that needs comments:
https://github.com/Marak/faker.js/pull/398

Marak:
@aexol -

Alternate solution could be using `faker.image.dataUri()` method

PR #856, opened 2019-12-15
Changes LoremPixel from HTTP to HTTPS

abnersajr:
## Description

It changes the protocol used by `lorempixel` to generate images.
This was discussed and requested in #652


Marak:
@abnersajr -

Thank you for your contribution. This is a smart change. Thank you for looking into issue #652 and creating this PR.

The one problem I can see is that certain users might need the option to load non-https images?

I'm going to merge for now and we can always add the option later.

PR #863, opened 2020-01-13
Add pt-br male & female first names

ArthNRick:


lib/locales/pt_BR/name/female_first_name.js
```diff
@@ -0,0 +1,52 @@
+module["exports"] = [
+         "Sophia",
```
Marak, line 2: Will merge if we can get the linting issue resolved here, spacing appears off.

Thank you!

Issue #851, opened 2019-12-01
Any way to filter out profanity?


vgjenks:
Generated an image URL w/ this value:

https://s3.amazonaws.com/uifaces/faces/twitter/fuck_you_two/128.jpg

All good and fine...unless you're generating fake data for clients to preview your APIs.

Seems like it'd be a good idea to be able to globally filter out any sort of profanity, or just not allow it through at all, to begin with?

Marak:
It looks like `fuck` is in the name of some of the free image avatars provided to the project. 

Not really an intentional addition.

If anyone wants to grep the codebase and remove the profanity I will merge the request.

PR #864, opened 2020-01-16
Remove profanity.

mastermatt:
Fixes https://github.com/Marak/faker.js/issues/851

@Marak I tried to run `npm run build` as well but it seems to be broken and not run for the last several changes. So this PR only includes the changes directly in the `lib` dir.


Marak:
Thanks for removing the profanity.

> seems to be broken and not run for the last several changes

I have no idea what this statement means. If you aren't able to build the project please post the error log in a new issue describing the problem. It's most likely your local environment.

PR #885, opened 2020-03-06
Change image url to stable provider

fjbarrett:
It could just be my ISP, but I have had a lot of trouble loading images from lorempixel.com. This changes the base url for the image request from lorempixel.com to placeimg.com, which share the same url structure when querying for random images.

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -47,7 +47,7 @@ var Image = function (faker) {
       if (typeof https !== 'undefined' && https === true) {
         protocol = 'https://';
       }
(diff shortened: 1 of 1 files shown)


Marak:
Thank you for your contribution @fjbarrett 

It's interesting, because I thought we had actually replaced the hard-coded lorempixel links with a provider pattern. Must have missed this one.

I took a quick look at the placeimg.com website and the company that runs it gomasuga.com. Seems OK to me.

Merged since this should just work. We can always circle back around and think about adding this as a new image provider / default image provider.

Issue #884, opened 2020-03-05
faker.finance.amount() returns a string


amartiuz:
Following the documentation it says: **Decimal number faker.finance.amount(min, max, decimal places) Random number between "min" and "max" including decimals to X digits.**

- **Expected behavior**: should return a **number**. ie: `122.12`
- **Current Behavior**: returns a **string**. ie: `"122.12"`

Marak:
Should be resolved with #887 and will be included in the next release.

Thank you taking the time to report this @amartiuz

PR #887, opened 2020-03-11
Fix return of finance.amount

edumoreira1506:
How @amartiuz said on issue #884 , the function `faker.finance.amount()` was returning a `string` instead of a `number`, sometimes this can cause a problem and don't make sense the function returns a string. This PR fixes this.

Behavior before of this changes: returns a string. ie: "122.12"
Behavior after of this changes: returns a number. ie: 122.12


Marak:
@edumoreira1506 

Thank you for your contribution. I appreciate you looking into @amartiuz issue and providing a solution.

It makes sense this should return a number. I'm trying to remember if there was a reason it didn't to begin with and cannot. Perhaps it was an oversight.

Going to merge for now and will do a double check on `finance.amount` before next release.

PR #891, opened 2020-03-15
Update README.md with ES6 syntax

raj7desai:
It does what it says. Updated `var` to `const` or `let` as needed


Marak:
All changes to the documentation must be done in `./build/src/docs.md`

Issue #883, opened 2020-03-05
[pt_BR] invalid URLs


gabrielmmontenegro:
Hi,

Thanks for the amazing work on this project.

I just wanna let you know that some of my tests are eventually breaking with invalid URLs when in ```pt_BR``` locale, which includes accents and spaces. 

Examples:

```
"https://júlio césar.info"
"http://vitória.br"
```

My project is running the latest version of faker: 4.1.0.

Best,
Gab

Marak:
Resolved with #895

PR #895, opened 2020-03-18
Fix special characters in internet.url

samir-araujo:
This PR addresses #883 

It adds a list of known special characters in `pt_BR` to be removed and replaces spaces with dashes to make a valid URL

### DIFF
--- a/lib/internet.js
+++ b/lib/internet.js
@@ -182,7 +182,7 @@ var Internet = function (faker) {
    * @method faker.internet.domainWord
    */
   self.domainWord = function () {
-      return faker.name.firstName().replace(/([\\~#&*{}/:<>?|\"'])/ig, '').toLowerCase();
+      return faker.name.firstName().replace(/([\\~#&*{}/:<>?|\"|àáãäâèéëêìíïîòóöôùúüûñç·/_,:;'])/ig, '').replace(/ /g, '-').toLowerCase();
   };
 
   self.domainWord.schema = {
--- a/lib/internet.js
+++ b/lib/internet.js
@@ -1,4 +1,5 @@
 var random_ua = require('../vendor/user-agent');
+var slugify = require('slugify');
 
 /**
  *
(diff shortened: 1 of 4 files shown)


lib/internet.js
```diff
@@ -182,7 +182,7 @@ var Internet = function (faker) {
    * @method faker.internet.domainWord
    */
   self.domainWord = function () {
-      return faker.name.firstName().replace(/([\\~#&*{}/:<>?|\"'])/ig, '').toLowerCase();
+      return faker.name.firstName().replace(/([\\~#&*{}/:<>?|\"|àáãäâèéëêìíïîòóöôùúüûñç·/_,:;'])/ig, '').replace(/ /g, '-').toLowerCase();
```
oscarmarcelo: This regex isn't enough. There are other languages with many other diacritical marks in many other base characters.

Perhaps using a Unicode normalizer would be wiser.

ES6 has [String.prototype.normalize()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize), and there are also packages like [remove-accents](https://www.npmjs.com/package/remove-accents) that although doesn't use native methods, but at least cover a lot more cases.

samir-araujo: Thanks for the reply, @oscarmarcelo!

It really wasn't enough.

After fiddling a little with a few libs, I found that [slugify](https://www.npmjs.com/package/slugify) probably works better, since it take care of special characters and diacriticals as well.

Also added a fallback case for languages like japanese that would return an empty string.

oscarmarcelo: LGTM! Nice work! 👏

lib/internet.js
```diff
@@ -1,4 +1,5 @@
 var random_ua = require('../vendor/user-agent');
+var slugify = require('slugify');
```
Marak, line 2: This is going to be a big no.

Marak:

`faker.js` does not use any third-party dependencies, so I cannot accept a PR which requires `slugify` library.

Closing for now. Feel free to open again if you can figure this out without adding third-party dependency.

Issue #842, opened 2019-10-16
faker.fake interpolation


polomoshnov:
`faker.date.between("2015-01-01", "2015-01-05")` works well but `faker.fake('{{date.between("2015-01-01", "2015-01-05")}}')` returns `Invalid Date`.

pedro-rates:
Do you guys accept pull requests? I've found a pretty easy way to fix this.

PR #896, opened 2020-03-19
Add multi parameters support for faker.fake

samir-araujo:
This PR addresses #842 

Since I couldn't find anything that transpile the code from es6 to es5, I used `fn.apply` with array instead of a `spread` to pass N parameters to the interpolated function.

Also to not break other implementations, I decided to add a `else if` validating if `params` is a string and has a comma in it.

### DIFF
--- a/lib/fake.js
+++ b/lib/fake.js
@@ -88,6 +88,8 @@ function Fake (faker) {
     var result;
     if (typeof params === "string" && params.length === 0) {
       result = fn.call(this);
+    } else if (typeof params === "string" && params.indexOf(',') !== -1) {
+      result = fn.apply(this, params.split(','));
     } else {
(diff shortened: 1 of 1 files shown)


Marak:
Thank you for your contribution @samir-araujo 

I *think* this looks good. I will review this along the other issues related to `faker.fake`

Tracking #916

caiotarifa:
Any update about this?

Marak:
I think the best course of action is going to be dropping the code used in `faker.fake` and simply rely on the string interpolation feature which is now available in JavaScript.

There is a more discussion about this on #916 

Does anyone have a compelling reason why we shouldn't remove `faker.fake` and replace it with regular string interpolation?

PR #897, opened 2020-03-27
Removed an unused variable, password, found in internet.js on line 354

Esquared722:


Marak:
Good catch @Esquared722 !

Merged.

PR #898, opened 2020-04-07
Removed redundant } in nl_BE city names

MrBramme:
Fixed an issue for the nl_BE locale. The City names get generated with a } between them if they're a compound of city_prefix and city_suffix.


Marak:
Good catch @MrBramme @andrewwippler 

Merged.

PR #908, opened 2020-05-05
Fix: Make faker.random.word return always only one word

### LINKED ISSUE #602: faker.random.words() returns phrases, not words

### LINKED ISSUE #661: Random word (singular) sometimes returns more than one word

bira37:
The method faker.random.word() can return multiple words sometimes (due to its behavior). To prevent that, I cut all spaces between the strings, joining them into one before returning.

Fixes #602 
Fixes #661

### DIFF
--- a/build/build/faker.js
+++ b/build/build/faker.js
@@ -69926,7 +69926,7 @@ function Random (faker, seed) {
 
     // randomly pick from the many faker methods that can generate words
     var randomWordMethod = faker.random.arrayElement(wordMethods);
-    return faker.fake('{{' + randomWordMethod + '}}');
+    return faker.fake('{{' + randomWordMethod + '}}').split(' ').join('');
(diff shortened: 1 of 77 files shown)


Marak:
@bira37 -

All Pull Requests to `faker.js` must be done *without* modification of the `/build` files. I cannot merge a 77 file change when only a few files have actually been changed.

Please feel to update the PR to only modify the files which are actually changed.

bira37:
Hi, @Marak 

I've rolled back the changes I made on /build folder. Now it should be ok.

Marak:
@bira37 

Same issue.

All Pull Requests to `faker.js` must be done without modification of any of the files that are built / generated. 

Please feel to update the PR to only modify the files which are actually changed.

PR #912, opened 2020-05-18
Fixes currency codes

Zache:
I double checked with https://www.iban.com/currency-codes


Marak:
@Zache -

Merged.

PR #913, opened 2020-05-28
clear unique

azai91:


Marak:
@azai91 -

Could you please include tests for this new method?

All `faker.js` methods require test coverage.

PR #914, opened 2020-05-30
Chore a

vamsi462:
you can writer the comments about  this change here.


Marak:
Incomplete. Tests failing. 

PR seems to be removing core `faker.js` methods.

Will not merge.

PR #924, opened 2020-06-24
Make user agent deterministic

starsprung:
User agent currently uses Math.random() and so produces different values for the same seed.


Marak:
@starsprung -

Is the change really this simple? Did we miss anything?

Marak:
Merged.

Issue #929, opened 2020-07-06
userAgent does not respect randomness seed


joshua-barnett:
On `faker@4.1.0`.

```
Welcome to Node.js v12.18.2.
Type ".help" for more information.
> const faker = require('faker')
undefined
> faker.seed(1)
undefined
> faker.internet.email()
'Zion.Reichel12@yahoo.com'
> faker.seed(1)
undefined
> faker.internet.email()
'Zion.Reichel12@yahoo.com'
> faker.seed(1)
undefined
> faker.internet.userAgent()
'Opera/11.67 (Windows NT 6.3; U; KG Presto/2.9.172 Version/11.00)'
> faker.seed(1)
undefined
> faker.internet.userAgent()
'Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 5.1; Trident/5.0)'
```

Marak:
I just checked on `master` and this appears to be working.

If you still think you are correct feel free to open a PR with a failing test.

Please try to take the time to investigate issues before reporting them here thank you.

PR #936, opened 2020-07-20
Add failing test for userAgent

joshua-barnett:
Proving it is not deterministic and does not respect random seed.

Previously reported and ignored.

- https://github.com/Marak/faker.js/issues/929

Issue appears to have been fixed by @starsprung in the following PR.

- https://github.com/Marak/faker.js/pull/924

Tested that this test is fixed by @starsprung's changes.

- https://github.com/joshua-barnett/faker.js/compare/bfe2c99...16cfc9f

### DIFF
--- a/test/internet.unit.js
+++ b/test/internet.unit.js
@@ -159,6 +159,14 @@ describe("internet.js", function () {
             var ua = faker.internet.userAgent();
             assert.ok(ua);
         });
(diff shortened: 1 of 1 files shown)


Marak:
Thank you for your contribution @joshua-barnett 

Merged.

PR #952, opened 2020-08-05
Finnish names

tuulahdus:
Added support for Finnish locale


Marak:
Thank you for your contribution @tuulahdus !

Usually I would ask for another native Finnish speaker to confirm the locality data, but this seems OK to me.

Please feel free to add more data!

Merged.

PR #953, opened 2020-08-06
Fix double BQ in country code for  #911

### LINKED ISSUE #911: BQ is in country_code twice
The ISO country code "BQ" is in the country_code.js list twice.

Moosh-be:
FIX  #911


Marak:
Thank you for addressing this issue @Moosh-be !

Much appreciated!

PR #962, opened 2020-08-12
Add method to get an object of two random elements

tuliooassis:
This PR provides a way to generate an object with two random elements.

It introduces the `faker.random.object()` method and outputs like `{grow: 'Investment Account', Principal: 'Lodge'}`.

Any feedback, I will be glad to know.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -135,6 +135,20 @@ function Random (faker, seed) {
   }
 
   /**
+   * object
+   *
+   * @method faker.random.object
+   */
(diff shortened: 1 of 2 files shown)


tuliooassis:
> You could add a `length` parameter to define the number of `key:value` in the object, with `length=2` as default.

Thanks @EdouardDem for the feedback.

tuliooassis:
Hi @Marak, do you have any update on this one?

Issue #960, opened 2020-08-12
faker.random.words(4) returns more then 4 words


RBee9:
`
import faker from 'faker'

for (let i = 0; i < 50; i++) {
    let word = faker.random.words(4).split(' ')
    if (word.length > 4) {
        throw Error(\`Found more then 4 word: "${word}"\`) 
    }
}
`

Because faker.random.word chooses a method to create the word randomly. Its possible that more words then suppled are generated.

RecepCil:
Actually, the method returns items as many as you wish, but some items have more words than one.


![resim](https://user-images.githubusercontent.com/18148117/90341110-8049cd00-e005-11ea-9ff1-1be9bf806ef2.png)

![resim](https://user-images.githubusercontent.com/18148117/90340593-fcdaac80-e001-11ea-8cb7-779438063f40.png)

Marak:
I think this has been solved in the latest release.

Please feel free to re-open with a failing test if you spot this again.

PR #965, opened 2020-08-16
random words generator - returning item count problem - bugfix

RecepCil:
I commented on #960


Marak:
Thank you for your contribution @RecepCil 

We cannot merge this PR in it's current form since it's making a modification to a build file, and not the core library.

Please feel free to re-open a new PR which only changes the core library itself.

PR #966, opened 2020-08-17
Suggestion to change the pt_BR location to make more sense in our country

vivianeflowt:
1- I added more common names from our country based on sites that have this data like this:

https://exame.com/brasil/os-100-nomes-mais-comuns-no-brasil-em-2014/
https://www.dicionariodenomesproprios.com.br/nomes-populares-comuns-no-brasil/

2 - City prefixes and suffixes do not make sense for cities in Brazil. This kind of thing is not used here, so I removed it.

3 - There is only one postal code standard in our country and I left only the correct one
 "#####-###"


lib/locales/pt_BR/address/city_prefix.js
```diff
@@ -1,7 +1,2 @@
 module["exports"] = [
-  "Nova",
```
Marak, line 2: We are we deleting this data?

lib/locales/pt_BR/address/city_suffix.js
```diff
@@ -1,6 +1,2 @@
 module["exports"] = [
-  "do Descoberto",
```
Marak, line 2: This data is not good?

leonidasv:
This change caused a bug: `undefined` is being concatenated to city name strings.

image

Marak:
@leonidasv -

Can you prove that with code? 

We can't be certain with the screenshot you posted as it has no context.

lib/locales/pt_BR/address/city_prefix.js
Marak, line 2: Deleting these probably wasn't the best idea.

Marak:
@leonidasv -

I was able to confirm.

If anyone can do a quick PR to add this data back that would be great.

Thanks for letting us know!

PR #968, opened 2020-08-18
Add Time Zone

mkg0:
Added `address.timeZone` with a simple test. Data was already there but the method was missing. Skipped doc generation because it was creating lots of noise(changes) in my machine.


Marak:
Good catch @mkg0 

Merged

PR #972, opened 2020-08-20
Fix typo in name inside manufacturer.js

nelsonjchen:


Marak:
Great catch @nelsonjchen !

PR #784, opened 2019-05-11
Migrate away from RawGit

Richienb:
Migrate demo URL from `rawgit.com` to `raw.githack.com`.

PR #978, opened 2020-08-25
Migrate away from RawGit

Richienb:
See #784 

// @Marak


Marak:
Thank you for your contribution. Getting off rawgit is probably a good idea.

We are not very interested in using third-party site to host content.

I'll try to setup a CI pipeline to marak.com as a free host.

Issue #973, opened 2020-08-21
Faker.random.word returns '24/365'


elkemper:
Hey!
I'm not sure, but it seems that slash char is not best candidate for "word", and it can cause failures, when random.word() is used for salting file names. It causes.

Marak:
@elkemper -

Are you able to reproduce this? Can you find the place in the code where these values are being returned?

elkemper:
@Marak 
This is inconsistent behavior, because this is random namespace. but, if you invoke Faker.random.word() for, i think 20000 times, you can find this value.

Marak:
@elkemper - 

Sounds good. 

Let us know what you can find. 

PRs are welcome.

elkemper:
@Marak 
Well, I've fixed only for EN locale, but, here it is
https://github.com/Marak/faker.js/pull/980

Marak:
@elkemper - Why not use `System.fileName()` method instead?

PR #980, opened 2020-08-27
[Issue #973] Delete 24/7 and 24/365 so it wouldn't appear in faker.random.word()

elkemper:
delete them from lib/locales/en/company/bs_adjective.js, as well as from lib/locales/en/company/descriptor.js (which is not included in random.word, but just in advance)


Marak:
I don't really see this as an issue.

If anyone can make a compelling case why we should remove this data please let us know.

Closing for now.

PR #985, opened 2020-08-28
perf(system): improve commonFileName, commonFileExt, fileType, directoryPath performance

### LINKED ISSUE #822: system.fileExt is slow
Hi, thanks for this great package.

I found some generators to be slow and looking at the source there might be ways to optimize them.

## `fileExt`

https://github.com/Marak/faker.js/blob/master/lib/system.js#L109

This generates the list of all mime extensions for each invocation. This can probably be memoized to avoid doing it again after a repeated call.
In my test it created a string array with >1k entries for each invocation.

This also means any generator, which consumes `fileExt`, e.g. `fileName` computes it again.

Are you open for PRs which try to tackle some performance issues?

makepanic:
fixes #822

### DIFF
--- a/lib/system.js
+++ b/lib/system.js
@@ -1,27 +1,72 @@
 // generates fake data for many computer systems properties
 
+var commonFileTypes = [
+  "video",
+  "audio",
+  "image",
+  "text",
+  "application"
+];
+
+var commonMimeTypes = [
+  "application/pdf",
+  "audio/mpeg",
+  "audio/wav",
+  "image/png",
+  "image/jpeg",
+  "image/gif",
+  "video/mp4",
+  "video/mpeg",
+  "text/html"
+];
+
+function setToArray(set) {
+  // shortcut if Array.from is available
+  if (Array.from) { return Array.from(set); }
+
+  var array = [];
+  set.forEach(function (item) {
+    array.push(item);
+  });
+  return array;
+}
+
 /**
  *
  * @namespace faker.system
  */
-function System (faker) {
+function System(faker) {
+  var typeSet = new Set();
+  var extensionSet = new Set();
+  var mimeTypes = faker.definitions.system.mimeTypes;
+
+  Object.keys(mimeTypes).forEach(function (m) {
+    var type = m.split("/")[0];
+
+    typeSet.add(type);
+
+    if (mimeTypes[m].extensions instanceof Array) {
+      mimeTypes[m].extensions.forEach(function (ext) {
+        extensionSet.add(ext);
+      });
+    }
+  });
+
+  var types = setToArray(typeSet);
+  var extensions = setToArray(extensionSet);
+  var mimeTypeKeys = Object.keys(faker.definitions.system.mimeTypes);
+
 
   /**
-   * generates a file name with extension or optional type
+   * generates a file name
    *
    * @method faker.system.fileName
-   * @param {string} ext
-   * @param {string} type
    */
-  this.fileName = function (ext, type) {
-    var str = faker.fake("{{random.words}}.{{system.fileExt}}");
-    str = str.replace(/ /g, '_');
-    str = str.replace(/\,/g, '_');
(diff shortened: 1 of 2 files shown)


makepanic:
This changes:
* order of definition initialization in the faker construction. This allows all fake methods to use the definitions in their constructors
* mime types and file types are now precomputed to an array to allow easy usage when picking on of its items
  * **Note**: initialization uses a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) which requires IE>= 11. I didn't find any browser compatibility matrix so this shouldn't really break things

I ran a quick (500 iterations) of each changed method through [tracerbench](https://www.tracerbench.com/) to see how it behaves. `Control` is the current master and `Experiment` is this change:

[artifact-2.pdf](https://github.com/Marak/faker.js/files/5141578/artifact-2.pdf)

It found that `commonFileName`, `commonFileExt`, `fileType` and `directoryPath` improved their runtime.

makepanic:
Seems like most of the execution time in the system methods is in `fileName`, `commonFileName` and `filePath`. I'll try to see how I can improve them more.

makepanic:
I made some adjustments to simplify the `fake` method implementation. This in turn improves any methods that use fake, e.g. company.companyName or company.catchPhrase

Updated tracerbench result:

[artifact-20.pdf](https://github.com/Marak/faker.js/files/5142141/artifact-20.pdf)

Marak:
Thank you for your contributions @makepanic 

We really appreciate you taking the time to track this down.

There seems to be two groups of optimizations going on here. It would be better if you could modify this PR to not make any changes to `Faker.fake` ( for now ). That way we can merge the changes to `System` module first.

Our thoughts are that the best optimization here might be replacing `faker.fake` to use string interpolation ( wasn't available when faker.js was created ). See: https://github.com/Marak/faker.js/pull/896

makepanic:
I've rebased the PR onto the current master.

I agree it makes sense to make this change as minimal as possible. I've removed everything outside the system changes.
The changes in index are required as it's using definitions which weren't defined yet on the previous versions.

Marak:
Thank you for your contributions @makepanic

Merged.

lib/system.js
```diff
@@ -1,45 +1,85 @@
...
+
 /**
  *
  * @namespace faker.system
  */
-function System (faker) {
+function System(faker) {
+  var typeSet = new Set();
+  var extensionSet = new Set();
+  var mimeTypes = faker.definitions.system.mimeTypes;
```
Marak, line 41: This seems no good.

When requiring locales individually using this syntax: `const Faker = require('faker/locale/en');`, `faker.definitions` will not yet be populated.

We'll have to revert this or figure out another solution.

RE: #1148

lib/system.js
```diff
@@ -1,45 +1,85 @@
...
-    str = str.replace(/\-/g, '_');
-    str = str.replace(/\\/g, '_');
-    str = str.replace(/\//g, '_');
-    str = str.toLowerCase();
-    return str;
+  this.fileName = function () {
+    var str = faker.random.words() + "." + faker.system.fileExt();
+    return str
+        .toLowerCase()
+        .replace(/\W/g, "_");
```
Marak, line 69: This is no good because it will also remove the `.` for the file extension.

@makepanic - Did you actually try using any of these methods or reviewing the actual test cases?

lib/system.js
```diff
@@ -1,45 +1,85 @@
...
-    str = str.replace(/ /g, '_');
-    str = str.replace(/\,/g, '_');
-    str = str.replace(/\-/g, '_');
-    str = str.replace(/\\/g, '_');
-    str = str.replace(/\//g, '_');
-    str = str.toLowerCase();
-    return str;
+    return str
+        .toLowerCase()
+        .replace(/\W/g, "_");
```
Marak, line 82: Same issue here. This will remove the `.` in the file extension and replace it with a `_`....

PR #989, opened 2020-09-01
Add optional country code parameter for iban method

Spea:
We have a few integration tests running where we need an IBAN from a specific country, but the `iban` method always returned a random one. 

This PR adds a second (optional) parameter to the `iban` method where you can pass the country from which you want to generate an IBAN

### DIFF
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -255,8 +255,14 @@ self.litecoinAddress = function () {
    * @method  faker.finance.iban
    */
   self.iban = function (formatted, countryCode) {
-      var findFormat = function(currentFormat) { return currentFormat.country === countryCode; };
-      var ibanFormat = countryCode ? ibanLib.formats.find(findFormat) : faker.random.arrayElement(ibanLib.formats);
+      var ibanFormat;
+      if (countryCode) {
+          var findFormat = function(currentFormat) { return currentFormat.country === countryCode; };
+          ibanFormat = ibanLib.formats.find(findFormat);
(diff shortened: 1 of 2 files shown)


lib/finance.js
```diff
@@ -248,10 +248,19 @@ self.litecoinAddress = function () {
...
+   * @param {string} [countryCode] - The country code from which you want to generate an IBAN, if none is provided a random country will be used.
+   * @throws Will throw an error if the passed country code is not supported.
+   *
    * @method  faker.finance.iban
    */
-  self.iban = function (formatted) {
-      var ibanFormat = faker.random.arrayElement(ibanLib.formats);
+  self.iban = function (formatted, countryCode) {
+      var findFormat = function(currentFormat) { return currentFormat.country === countryCode; };
+      var ibanFormat = countryCode ? ibanLib.formats.find(findFormat) : faker.random.arrayElement(ibanLib.formats);
```
Marak, line 259: Could you please remove the conditional assignment operator here?

We don't allow these and we are in the process of adding a linting rule that forbids this.

Marak:
@Spea -

I've made a code comment about the conditional assignment operator. If you can address this we can merge the PR.

Thanks again!

lib/finance.js
Spea: I adjusted it accordingly. Hope it is ok like that.

Marak:
Thank you for your contributions @Spea 

Merged!

PR #990, opened 2020-09-04
add korean, thailander name list

0x0a0d:
+ Thailand name with gender
+ Korea name


lib/index.js
```diff
@@ -130,7 +130,7 @@ function Faker (opts) {
     _definitions[d].forEach(function(p){
       Object.defineProperty(self.definitions[d], p, {
         get: function () {
-          if (typeof self.locales[self.locale][d] === "undefined" || typeof self.locales[self.locale][d][p] === "undefined") {
+          if (typeof self.locales[self.locale] === "undefined" || typeof self.locales[self.locale][d] === "undefined" || typeof self.locales[self.locale][d][p] === "undefined") {
```
Marak, line 133: Why did this line need to change?

Marak:
@0x0a0d -

This looks good!

Usually with adding new locales we don't need to modify the core library? I made a code comment. If you can please address this issue we should be able to merge this.

lib/index.js
0x0a0d: fix this
When I accidentally call faker.setLocale('something-does-not-exists'), faker crash

Marak:
@0x0a0d -

I'm trying to understand which locales these are: `jp_katakana`, `jp_hiragana`, `jp_kanji`

Are these official i18n localities? `faker` usually only supports actual `i18n` locales.

Here is an example list: https://github.com/ladjs/i18n-locales

Can you explain a bit what is being added here? Are these not just part of `jp` and `jp_JP`?

dungsil:
`kr` or `jp` is not language code.

`kr` -> `ko`
`jp` -> `ja`

Hiragana, katakana, and Kanji are just Japanese.

Issue #984, opened 2020-08-28
`finance.amount` sometimes returns value with only one decimal place


Marak:
```
faker.finance.amount()
faker.finance.amount()
faker.finance.amount()
```

```
95.15
669.55
550.8
```

That last one, `550.8` should probably read `550.80` instead.

Would appreciate if anyone could sort this out at:

```
finance.unit.js
  amount(min, max, dec, symbol)
```

mastermatt:
Based on this PR https://github.com/Marak/faker.js/pull/887, `faker.finance.amount()` returns a number instead of a string now. 
So a one digit decimal would be expected in the case of `550.80`.

Note, the JSDoc still says it returns a string. Given the option to pass in a currency symbol, casting the return value to a number seems like the wrong direction.

Marak:
@mastermatt -

Seems like we might have to revert the change for #887 ?

If anyone can get a PR going I will merge.

king-11:
@Marak i would like to work on this issue can i go ahead

Marak:
I always get a little confused when people ask me if they can work on an issue.

The answer is always yes @king-11 

Permission is not required.

king-11:
what i found is that as per ECAMScript specification for Number in any sort of conversion from string to number they use [this](https://tc39.es/ecma262/#sec-tonumeric) specfication where addition is taking place in  StrUnsignedDecimalLiteral:::DecimalDigits.DecimalDigits  part so they don't take into account the zeros. So i guess it won't be possible to have trailing non-significant zeros in the number as per ECMAScript specification.

Marak:
Resolved with #994 🎉

PR #994, opened 2020-09-06
Fix return of finance.amount to string

Lukazovic:
On issue #984, it was said that the function faker.finance.amount() was returning a number instead of a string. This PR fixes this.

Behavior before of this changes: returns a number. ie: 
```js
faker.finance.amount()
faker.finance.amount()
faker.finance.amount()

95.15
669.55
550.8
```

Behavior after of this changes: returns a string. ie:
```js
faker.finance.amount()
faker.finance.amount()
faker.finance.amount()

"95.15"
"669.55"
"550.80"
```

### DIFF
--- a/lib/finance.js
+++ b/lib/finance.js
@@ -112,7 +112,7 @@ var Finance = function (faker) {
       var randValue = faker.random.number({ max: max, min: min, precision: Math.pow(10, -dec) });
       var stringNumber = symbol + randValue.toFixed(dec);
 
-      return Number(stringNumber);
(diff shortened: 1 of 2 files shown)


Marak (approved):

Marak:
Thank you @Lukazovic , this really helps!

It's easy for us to let the little things slip through sometimes.

We really appreciate to see small PRs like this.

rtymchyk:
@Marak any reason there hasn't been a new release to capture this fix?

PR #995, opened 2020-09-07
adds optional parameters to .url, docs, and tests

jgrossbergpear:
This PR adds optional parameters to `internet.url()`, inspired by similar functionality in [the ruby `faker` gem](https://www.rubydoc.info/github/stympy/faker/Faker%2FInternet.url)

This is convenient, for example, when:
* wanting to generate only secure URLs, with randomized domain names
* wanting to generate only insecure (HTTP) URLs, with randomized domain names
* wanting to generate a mix of HTTPS and HTTP urls, all with the same domain name

### DIFF
--- a/doc/faker.internet.html
+++ b/doc/faker.internet.html
@@ -33,7 +33,7 @@
 			<li class="dropdown">
 				<a href="namespaces.list.html" class="dropdown-toggle" data-toggle="dropdown">Namespaces<b class="caret"></b></a>
 				<ul class="dropdown-menu ">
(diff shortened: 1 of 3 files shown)


Marak (changes requested):
Changes requested.

jgrossbergpear:
Hello @Marak -- thank you for taking the time to review my PR.

I see you requested changes, but I cannot see what those requested changes *are*.

Can you clarify what needs changing? Thanks!

PR #998, opened 2020-09-09
npm package has 37MB of coverage reports published in it

mhart:


mhart:
Also, is there any reason this package doesn't just distribute `dist` as the npm package? Or even just `dist/faker.min.js`? It seems to have everything in it. Everything else seems redundant?

ohcibi:
@mhart we are currently actively working on a new release which will cover that amongst other legacy stuff (eg well get rid of the fake function and it’s dependencies in favor of native string interpolation). The current structure is a result of the i18n system this library has established. We are aware of the problematic implications.
We want to address as much critical pull requests that are not related to the architecture or the module structure as possible. Well let ya all know when we have our first sneak peak ready.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

mhart:
Thanks @ohcibi – in the meantime I've just forked faker (just `en`) over at https://github.com/bustle/faker.js – it's only 960kb to install 👍

Marak:
This has been resolved with a few other changes to `.npmignore` found https://github.com/Marak/faker.js/blob/master/.npmignore

Closing this since we've gotten the package size much smaller now.

We are tracking additional issues to reduce size / provide incremental locale builds.

PR #1010, opened 2020-09-24
fix: reduce size of the npm package

yhnavein:
Updated `.npmignore` file to include files that are not necessary in the bundle.

Package size difference:

**Before**:
```
npm notice package size:  6.1 MB
npm notice unpacked size: 43.5 MB
npm notice total files:   4692
```

**After**:
```
npm notice package size:  1.9 MB
npm notice unpacked size: 8.4 MB
npm notice total files:   1585
```

This should not affect anyone in a negative way and for everyone it will improve install times :)

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -8,3 +8,15 @@ examples/
 meteor/
 test/
 logo.png
+.DS_Store
+reports/
+.github/
+.nyc_output/
+*.yml
+.versions
+*.png
+*.lock
+.jshint*
+gulpfile.js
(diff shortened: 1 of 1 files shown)


mxcoder:
I actually got a disk space warning out of this. Please merge and publish asap.

ssilve1989:
These unnecessary files also create thousands of symlinks when used with the Bazel build system.

Marak:

We've resolved this with other changes to the `.npmignore`.

PR #1011, opened 2020-09-30
Allow clean found items after an specific value

roggervalf:
We could optionally pass a value to clean the found items, so we could maintain a number of unique values and after an specific number of found items we could clean it, so it is ok to have a duplicate on a new group of values.

### DIFF
--- a/test/unique.unit.js
+++ b/test/unique.unit.js
@@ -1,54 +1,71 @@
-if (typeof module !== 'undefined') {
-    var assert = require('assert');
-    var sinon = require('sinon');
-    var faker = require('../index');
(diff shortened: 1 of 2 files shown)


Marak:
@roggervalf -

Thank you for your contributions to `unique` and `clear` functionality.

We can see that this functionality should be added, but still want to do additional tests and review of implementation locally.

I'm going to take some time before the next major release and review all `unique` code and functionality and get this resolved.

RE: 
 - #1022
 - #913

roggervalf:
sure, no problem, I would recommend to test #1022 first and later I could resolve merge conflicts in this PR. Thanks for your revision

Issue #767, opened 2019-03-27
Add comma to thousands for faker.finance.amount 


chrisyap:
Is there a way to return 6,234.32 with a comma when using finance amount?
faker.finance.amount(6000, 6400, 2)

sbaechler:
The formatting is locale specific.

You can use `number.toLocaleString()`. (Requires Node > 9 with i18n extension.)

https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

Marak:
It might be nice if we could add this as a simple argument `autoformat` to `finance.amount`.

strobelt:
Can I take care of this? Seems like a simple enough change to the `finance.amount` method

Marak:
Thanks for this information @chrisocast @schlpbch 

Thank you for your contributions @strobelt.

This should now be resolved on `master` with #1015 

Closing.

PR #1015, opened 2020-10-02
Add autoFormat flag on finance amount

strobelt:
Motivated by this issue #767 the autoFormat flag was created on finance.amount().
When `true` this flag will format the generated finance amount on the current locale.

Example:
```
const amount = faker.finance.amount(1000, 1000, 2, "$", true); // on a en-US locale
console.log(amount); // $ 1,000.00


Marak:
Thank you for your contributions @strobelt !

This looks OK to me.

Merging.

Issue #920, opened 2020-06-14
Seeding 0 does not work


tomdohnal:
When I use the `seed` method like this: 

```javascript
faker.seed(0); // seeding 0
```

And then call *any* function which generates random results, I don't get the same results in for the same calling order. Other values besides 0 work fine, so it might be that you're checking in the `seed` value is "falsy" (which 0 is) thus resulting in not seeding whatsoever.

Marak:
I'm guessing this might have to do with the implementation in https://github.com/Marak/faker.js/blob/master/vendor/mersenne.js?

I'm not sure if `0` not working as a random seed is intended behavior or a bug.

Either way, `0` not working as a seed is unexpected behavior to the user so we should resolve it.

The easiest solution might just be not allowing `0` and adding a simple unit test?

**Will accept a PR for this.**

josecarlosweb:
The Mersennne Twistter algorithm consider seeder = 0 like nothing.

strobelt:
I found that the problem was the seed validation on `lib/random.js`. It was checking with `if(seed)` which doesn't work for 0 and can have many false positives for variables that aren't numbers. I changed the check and everything worked. I am just finishing a PR and will send it shortly. 😊

PR #1021, opened 2020-10-04
Allow zero as random seed

strobelt:
Fix bug reported on issue #920 that prevented seeding faker.random with a 0 and tests seeding with 0


Marak:
Thank you for your contribution @strobelt 

We appreciate you tracking this down.

Merged.

Resolves: #920

PR #1022, opened 2020-10-04
Scope in unique function and add clear function for found items

roggervalf:
This PR is trying to refactor unique function to allow use it with an scope, also adding a clear function when you want to reset the found items. Hope this could be useful, if there is something I should change let me know please.

PD: sorry about the lint changes.


roggervalf:
@Marak could you please review this PR?

PR #1026, opened 2020-10-10
- Added a new locale en_KE

Jeffrey-Kimani:


Marak:
@Jeffrey-Kimani -

`faker` has a fallback system in place for locality data so we don't have to duplicate existing data. We only need to add additional data for the specific locale for `en_KE` and then `en` will work for everything else.

Can you modify this PR to only include the data unique to `en_KE`? Thanks!

Jeffrey-Kimani:
Okay, let me check what i can strip down and submit another PR

Jeffrey-Kimani:
Done. I have cleaned up the code and removed unnecessary files

Issue #1025, opened 2020-10-09
random.word sometimes returns with parenthesis


hsw107:
We are using `faker.random.word()` inside some regex for testing purposes, and have been using `faker.random.word()`. 

It appears that there's a chance faker can grab a malformed word.
1. Faker selects a random item from the sources such as `Slovakia (Slovak Republic)` (this one is from `address.country`)
2. Faker splits this by space `["Slovakia", "(Slovak", "Republic)"]`
3. Faker returns out `"(Slovak`
When this gets used inside of regex, it throws an error.
`Invalid regular expression: /(Slovak/: Unterminated group`

We can of course just not use random word, and instead pull from a more specific source that doesn't include this punctation. But it would make sense to me that faker shouldn't be returning the open parenthesis here.

seanmetzgar:
Hi, I addressed this in my PR #1028

PR #1028, opened 2020-10-15
Ensure random word doesn't contain unwanted chars

seanmetzgar:
- Adds regex to `random.word` to trim unwanted characters (Issue #1025)
- Adds `package-lock.json` to .gitignore file


.gitignore
```diff
@@ -17,6 +17,7 @@ reports/
 /npm-debug.log
 /coverage
 /.npm
+package-lock.json
```
hsw107, line 20: This should not be in the gitignore

lib/random.js
```diff
@@ -203,7 +205,10 @@ function Random (faker, seed) {
     // randomly pick from the many faker methods that can generate words
     var randomWordMethod = faker.random.arrayElement(wordMethods);
     var result = faker.fake('{{' + randomWordMethod + '}}');
-    return faker.random.arrayElement(result.split(' '));
+    result = faker.random.arrayElement(result.split(' '));
+
+    // sanitize and retun word
```
hsw107, line 210: Typo on "return"

hsw107 (commented):
Thanks for doing this! Left some feedback that should hopefully help your PR go through!

.gitignore
seanmetzgar: thanks, updated this with my latest commit 😄

lib/random.js
seanmetzgar: thanks, updated this with my latest commit 😄

.gitignore
ohcibi: While in general this is true we haven't committed that file to master yet either. I think it's best to remove it from this PR. We will address the Lockfile issue in a separate branch.

/cc @Marak

hsw107: @ohcibi So the package-lock.json should be removed from the PR, but it still should **not** be added to the .gitignore, is that correct?

seanmetzgar: No problem, @ohcibi & @hsw107, I have removed the package-lock.json file in my latest commit, and left it out of .gitignore.

ohcibi: @hsw107 Let's wait for @Marak 's response. Im not sure. Maybe it is to not force anybody to either yarn or npm. I'll update the readme on this if necessary.

A general rule of thumb Im following myself is if in doubt I don't change anything unrelated to my pull request's topic.

lib/random.js
```diff
@@ -166,6 +166,8 @@ function Random (faker, seed) {
    * @param {string} type
    */
   this.word = function randomWord (type) {
+    // regex statement used to remove unwanted characters from beginning/end of words
+    var wordSanitizerRegex = /^[\s\(\)\[\]\{\}\.,\-'"]+|[\s\(\)\[\]\{\}\.,\-'"]+$/g;
```
Moosh-be, line 170: This list is perhaps "english centrics".
Is it difficult (or usefull or possible or untimely or premature ) to get  the list in locale config ?

hsw107: Is there something about this list that is english centric?
I'm not sure what a locale config of making something more regex-friendly would look like.

PR #1041, opened 2020-11-03
Bug Fix: Corrected reference to af_ZA & zu_ZA.

rhyswilliamsza:
Hi There

A simple PR. Seems like `af_ZA` & `zu_ZA` weren't linked correctly through `locales.js`.

Sidenote: it seems a little weird having these split out into three locales (firstly as it means phone number data etc is repeated, and secondly it all references a single country and thus should have a single point of truth) I would like to propose joining them into a single `za` locale. This is done in `java-faker` under `en_ZA`. This shouldn't be a breaking change at this point, as `af_ZA` and `zu_ZA` weren't in use anyways ;)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -16,6 +16,7 @@
 ## Usage
 
 ### Browser
+
 ```html
     <script src = "faker.js" type = "text/javascript"></script>
     <script>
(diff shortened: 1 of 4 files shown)


Marak:
@rhyswilliamsza -

Thank you for your contribution and looking into this.

I agree. Could you make a new PR and consolidate these localities into `en_ZA` and remove any duplicate data? That would really help us.

Can you remove the `README` from this PR? We don't modify that directly. Thanks!

PR #1043, opened 2020-11-04
Added support for random number generation with a fixed digit length

vanhooferwin:
Added support for generating numbers with a fixed digit length (allowed digit length from 1 to 15)

`Faker.random.number({digits: 1});` -> between 1 and 9
`Faker.random.number({digits: 5});` -> between 10000 and 99999

When digits option is set min, max and precision are ignored.

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -15,10 +15,10 @@ function Random (faker, seed) {
     }
   }
   /**
-   * returns a single random number based on a max number or range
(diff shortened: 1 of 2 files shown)


tobyhinloopen:
`options.digits > 0 && options.digits < 15`

If you provide `digits: 16`, it does something differently without any error, which is really unexpected.

Maybe convert it to a `random.digits` instead, that returns digits as a string (incl. leading zeroes)

PR #1053, opened 2020-11-18
Generate random vehicle vrm (vehicle registration mark)

andrewmcoupe:
This change allows the generation of random vehicle VRM's typically found in the UK in the format MF69RFT.


andrewmcoupe:
@Marak - any idea why litecoinAddress() unit test would fail after adding a new method to vehicle and going nowhere near the finance code?

roggervalf:
> @Marak - any idea why litecoinAddress() unit test would fail after adding a new method to vehicle and going nowhere near the finance code? I've done some digging but cannot understand why this would happen.

hi @andrewmcoupe, could you try to recover the package-lock.json from master, maybe it could help.

andrewmcoupe:
> > @Marak - any idea why litecoinAddress() unit test would fail after adding a new method to vehicle and going nowhere near the finance code? I've done some digging but cannot understand why this would happen.
> 
> hi @andrewmcoupe, could you try to recover the package-lock.json from master, maybe it could help.

@roggervalf Yeah I've tried that but that 1 test is still broken, strange 🤷‍♂️

roggervalf:
Hey @andrewmcoupe, I took a look in this issue, it is about the finance test:
https://github.com/Marak/faker.js/blob/master/test/finance.unit.js#L262
This regex expression should be changed to `/^[LM3][1-9a-km-zA-HJ-NP-Z]{25,32}$/`, so you can replace the line with:
```js
assert.ok(litecoinAddress.match(/^[LM3][1-9a-km-zA-HJ-NP-Z]{25,32}$/));
```
This test has an issue related to the numbers because it is expected that the returned expression has between 26-32 characters, but in this test as we have the first part with [LM3] at the beginning, this count as 1 character, so [1-9a-km-zA-HJ-NP-Z] should be repeated to 25-32 times. Hope I could help you with this, also I think that you can delete your package-lock too, since this file is not included

andrewmcoupe:
> Hey @andrewmcoupe, I took a look in this issue, it is about the finance test:
> https://github.com/Marak/faker.js/blob/master/test/finance.unit.js#L262
> This regex expression should be changed to `/^[LM3][1-9a-km-zA-HJ-NP-Z]{25,32}$/`, so you can replace the line with:
> 
> ```js
> assert.ok(litecoinAddress.match(/^[LM3][1-9a-km-zA-HJ-NP-Z]{25,32}$/));
> ```
> 
> That error was thrown because there is a faker.seed, since there is a new test related to your changes, this affected the result on this test and this test has an issue related to the numbers because it is expected that the returned expression has between 26-32 characters, but in this test as we have the first part with `[LM3]` at the beginning, this count as 1 character, so `[1-9a-km-zA-HJ-NP-Z]` should be repeated to 25-32 times. Hope I could help you with this, also I think that you can delete your package-lock too, since this file is not included

Ah, great work @roggervalf! Fixed the test, removed package-lock.json and pushed changes 👍

andrewmcoupe:
How often do PRs get reviewed? Seems like a lot of PRs have been open for a long time!

roggervalf:
Seems we it is related to #1046

roggervalf (approved):

Marak:
Thank you for your contributions @andrewmcoupe and @roggervalf 

Merged!

PR #1059, opened 2020-11-27
Prevent all locales loading if the only one is used

Taras-Verveha:
Current faker instance passed into user-agent generate function to avoid loading additional locales


Marak:
Thank you for your contribution @Taras-Verveha 

Merged.

PR #1064, opened 2020-12-02
Fix 663

FabianTe:
Hi, I'd like to get the fix that was originally introduced PR #792 going.

I hope I don't offend anyone by opening a new PR. I could have also created a PR into the branch of #792 but I think this is less hastle.

---

**Changes**

- Add `faker.address.buildingNumber`
- Add `helpers.templateString`
- Add street properties to `address` definition

FYI @PeachIceTea


FabianTe:
@Marak Sorry to bother. I have been looking at the CI quiet a while and replicated the issue locally.

I can't figure out why `npm run test` fails with the memory allocation error. I guess there is some kind of recursion or similar problem. On `master` the tests work fine (tested locally), on this branch not. Since this PR does not change anything related to the actual tests, am I correctly assuming that this is some kind of follow up problem of the address tests not passing (since the implementation was changed but not the tests)? Could you give me a pointer?

lib/address.js
```diff
@@ -125,22 +132,23 @@ function Address (faker) {
...
-          address = Helpers.replaceSymbolWithNumber("####") +  " " + faker.address.streetName();
-          break;
-      case 2:
-          address = Helpers.replaceSymbolWithNumber("###") + " " + faker.address.streetName();
-          break;
-      }
+      var address = Helpers.templateString(faker.definitions.address.street_address[0], {
+        street_name: faker.address.streetName(),
+        building_number: faker.address.buildingNumber()
+      })
```
FabianTe, line 138: This breaks the tests. Not all locales return an address that start with a number but that is what the tests assume. I guess the tests have to be changed somehow to reflect that? Unsure how to change them though...

FabianTe: I think the ` occasionally returns a X-digit street number` tests don't make sense. I have looked at some locales and there are multiple `building_number.js` templates that don't contain templates for 3, 4 or 5 sigit house numbers.

These tests fail now because only now the street address is properly build from the templates. Therefore I think the tests should be removed or replaced. Given that the current implementation of the address is kind of unusable IMO I'd prefer deleting the tests instead of coming up with a set of complex tests.

Marak:
@FabianTe -

We can't merge this Pull Request with failing tests.

We think the best thing to do here is modify the existing tests to cover the new code changes. It's important we test all the code paths. Feel free to make the necessary modifications to the test file and we can get this merged.

Thanks again!

lib/helpers.js
```diff
@@ -214,6 +214,24 @@ var Helpers = function (faker) {
    * @param {object} data
    */
   self.mustache = function (str, data) {
+    if (typeof str !== 'string') {
```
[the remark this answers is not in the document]
nahakiole: I tried debugging it with the Chrome Web Debug Tools attached to the Node Process, but I didn't get any information why the error appears. 
https://nodejs.org/en/docs/guides/debugging-getting-started/

PR #1067, opened 2020-12-04
Include only required files in npm archive

### LINKED ISSUE #1049: Coverage reports are included in npm archive
Roughly 30MB of coverage reports are included in the npm archive when running `npm install faker`. Not sure they should be included. It also looks like there is some other stuff that doesn't need to be in the npm archive.

Package Version: 5.1.0

HackbrettXXX:
Use "whitelist" instead of "blacklist"

Fix #1049.


HackbrettXXX:
Solved on master.

PR #1068, opened 2020-12-04
Possible solution for uifaces.com cancelled

caiotarifa:
https://uifaces.com


MrToto54:
We can find an alternative of https://uifaces.com at https://uifaces.co/
It can be interesting to use it.

caiotarifa:
@MrToto54 Working with UiFaces.com needs external API access. RandomUser.me allows you to use the current architecture, which will cause less compatibility problems

crismarucci (approved):

crismarucci:
Still not working cause of uifaces.com cancellation, I think that this solution is ready to go.

Marak:
This has already been resolved in `master` and the latest published version.

Please feel free to update to the latest and see if avatars are working.

Cheers.

PR #1077, opened 2020-12-25
Create female_first_name.js

navece:
adding some female first name in the list


navece:
after adding both female and male lists we can import it in index.js
so please apply my changes so i can add lines in index

Issue #1069, opened 2020-12-07
Incorrect name generation when specifying the gender using the Spanish locale.


CristianH033:
Hello.
When I set the Spanish locale 'es' the name generation works correctly using `faker.name.firstName()`:
```js
// data.js
var faker = require('faker');

faker.locale = "es";

var primerNombre = faker.name.firstName();
var segundoNombre = faker.name.firstName();

console.log(primerNombre);
console.log(segundoNombre);
```
Results:

```sh
❯ node data.js
Laura
Nicolás
❯ node data.js
Amalia
Vicente
❯ node data.js
Mariana
Estela
❯ node data.js
David
Eduardo
```

But when I specify the genre using `faker.name.firstName(1)`, the selected locale stops working and the generated names are English names:

```js
// data.js
var faker = require('faker');

faker.locale = "es";

var primerNombre = faker.name.firstName(1);
var segundoNombre = faker.name.firstName(1);

console.log(primerNombre);
console.log(segundoNombre);
```
Results:

```sh
❯ node data.js
Sadie
Alice
❯ node data.js
Samantha
Jacquelyn
❯ node data.js
Marcella
Maryann
❯ node data.js
Paulette
Laura
❯ node data.js
Lena
Wendy
```

navece:
@CristianH033 I want to work on that please assign me

Kenta1561:
As far as I know, the Spanish locale among others have no separation between male/female first names in the underlying data array. So in your case, because you specified the gender, it falls back to the default locale which is English. Correct me if I'm wrong.

navece:
@Kenta1561 exactly
But i have requested a pull to solve that issue with some code

Marak:
Thanks for letting us know @CristianH033 

We've added gendered `es` names and this should be partially resolved with #1080

Please checkout `master` to review. If anyone wants to add more names to the list please create a new PR.

PR #1080, opened 2020-12-26
issue #1069

navece:
faker.locale = "es";
console.log(faker.name.firstname(1)); //this will work now


Marak:
Thank you for your contribution @navece

Merged.

Could you do a follow-up Pull Request which contains a few more gendered `es` names? The current list for `es` is now short.

Thanks again!

PR #1083, opened 2020-12-30
hexaDecimal() fix: lowercase letters only and custom prefix

maximki:
`random.hexaDecimal` changes:
- removed uppercase letters for consistency. Having 'abcdef' and 'ABCDEF' was doubling the chance of getting a letter in each position as well
- added an option to have a custom (or empty) prefix. There is no need in `0x` in some cases
- updated tests

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -296,18 +296,22 @@ function Random (faker, seed) {
    *
    * @method faker.random.hexaDecimal
    * @param {number} count defaults to 1
+   * @prefix {string} prefix defaults to 0x
    */
-  this.hexaDecimal = function hexaDecimal(count) {
+  this.hexaDecimal = function hexaDecimal(count, prefix) {
     if (typeof count === "undefined") {
       count = 1;
     }
-
-    var wholeString = "";
-    for(var i = 0; i < count; i++) {
-      wholeString += faker.random.arrayElement(["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f", "A", "B", "C", "D", "E", "F"]);
+    if (typeof prefix === "undefined") {
+      prefix = "0x";
     }
 
-    return "0x"+wholeString;
+    var wholeString = prefix;
+    for(var i = 0; i < count; i++) {
+      wholeString += faker.random.arrayElement(["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f"]);
+    }
+
+    return wholeString;
   };
 
   return this;
--- a/test/random.unit.js
+++ b/test/random.unit.js
@@ -297,7 +297,22 @@ describe("random.js", function () {
 
     it('should generate a random hex string', function() {
       var hex = hexaDecimal(5);
-      assert.ok(hex.match(/^(0x)[0-9a-f]+$/i));
+      assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
+    })
+
+    it('should generate a random hex string with no prefix', function() {
+      var hex = hexaDecimal(5, "");
+      assert.ok(hex.match(/^[0-9a-f]{5}$/i));
+    })
+
+    it('should generate a random hex string with a custom prefix', function() {
+      var hex = hexaDecimal(5, "hex");
+      assert.ok(hex.match(/^(hex)[0-9a-f]{5}$/i));
+    })
+
+    it('should generate a random hex string with lowercase letters and numbers only', function() {
+      var hex = hexaDecimal(500, "");
+      assert.ok(hex.match(/^[0-9a-f]{500}$/i));
     })
   })
 


Marak:
Thank you for your contribution @maximki 

Wouldn't uppercase letters be preferred?

Can anyone else comment on this Pull Request and see if we've gotten this right?

maximki:
@Marak thanks for the feedback.
I might be wrong but it feels like many programming languages I dealt with use lowercase letters in hex numbers. Lowercase letters are also used in other parts of random.js. UUID generation, for example.
I just wanted consistency, so I'm happy to switch it to uppercase if required

Marak:
Can anyone else comment on this?

 Does anyone oppose to this change?

vnq:
Just noticed this pull request which is similar to mine #1124. I also replaced some duplicate code.

About the lowercase/uppercase, this is what [wikipedia](https://en.wikipedia.org/wiki/Hexadecimal) states:

> There is no universal convention to use lowercase or uppercase, so each is prevalent or preferred in particular environments by community standards or convention; even mixed case is often used.

Perhaps we could add 3 options:
* lower
* upper
* mixed

Unsure how you would like to implement those options though.

maximki:
@vnq all other methods in random.js use lowercase letters by default. The change is to keep consistency with lowercase letters through methods in random.js. If a user wants to have letters in uppercase they can use toUpperCase().

The method doesn't guarantee that there will be any letters in the returned string. So the method can't be used to properly test somebody's hex parsing code. Because of that I don't see a reason to add an option to select the case (lower, upper, mixed). The method should just return a correct hex number with/without a prefix. The change is just about consistency (and a custom prefix when needed).

vnq:
@maximki lowercase only is fine by me. Hence the perhaps.

You could however remove the case-insensitive flag from the tests, since we know they should be lowercase.
And fix the duplicate code in `git.js`, `internet.js` see #1124.

I was wondering about the `prefix` though, I made it a boolean in my pull request. As I'm unsure what other prefix you would need. Next thing somebody will be adding a suffix as well...

In my opinion for v6 we could remove the prefix from `faker.random.hexaDecimal` and just add it to the only function that uses this at the moment being:
```
  self.ethereumAddress = function () {
    var address = faker.random.hexaDecimal(40);

    return "0x"+address;
```

maximki:
hexaDecimal() is used in client code. Removing the prefix by default will break client code

I like your idea of making `prefix` a boolean flag. I'll add it
I will update other files like `git.js`, `internet.js`. Thanks for the suggestion

test/datatype.unit.js
```diff
@@ -234,9 +234,21 @@ describe("datatype.js", function () {
             assert.ok(hex.match(/^(0x)[0-9a-f]{1}$/i));
         });
 
-        it('generates a random hex string', function () {
+        it('generates a random hex string with prefix', function() {
             var hex = hexaDecimal(5);
-            assert.ok(hex.match(/^(0x)[0-9a-f]+$/i));
+            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
```
vnq, line 239:
```suggestion
            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/));

```

test/datatype.unit.js
```diff
@@ -234,9 +234,21 @@ describe("datatype.js", function () {
             assert.ok(hex.match(/^(0x)[0-9a-f]{1}$/i));
         });
 
-        it('generates a random hex string', function () {
+        it('generates a random hex string with prefix', function() {
             var hex = hexaDecimal(5);
-            assert.ok(hex.match(/^(0x)[0-9a-f]+$/i));
+            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
+            hex = hexaDecimal(5, true);
+            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
```
vnq, line 241:
```suggestion
            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/));

```

test/datatype.unit.js
```diff
@@ -234,9 +234,21 @@ describe("datatype.js", function () {
...
             var hex = hexaDecimal(5);
-            assert.ok(hex.match(/^(0x)[0-9a-f]+$/i));
+            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
+            hex = hexaDecimal(5, true);
+            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
+        });
+
+        it('generates a random hex string with no prefix', function() {
+            var hex = hexaDecimal(5, false);
+            assert.ok(hex.match(/^[0-9a-f]{5}$/i));
```
vnq, line 246:
```suggestion
            assert.ok(hex.match(/^[0-9a-f]{5}$/));

```

test/datatype.unit.js
```diff
@@ -234,9 +234,21 @@ describe("datatype.js", function () {
...
+        });
+
+        it('generates a random hex string with lowercase letters and numbers only', function() {
+            var hex = hexaDecimal(5000, false);
+            assert.ok(hex.match(/^[0-9a-f]{5000}$/i));
```
vnq, line 251:
```suggestion
            assert.ok(hex.match(/^[0-9a-f]{5000}$/));

```

test/datatype.unit.js
```diff
@@ -234,9 +234,21 @@ describe("datatype.js", function () {
...
-        it('generates a random hex string', function () {
+        it('generates a random hex string with prefix', function() {
             var hex = hexaDecimal(5);
-            assert.ok(hex.match(/^(0x)[0-9a-f]+$/i));
+            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
+            hex = hexaDecimal(5, true);
+            assert.ok(hex.match(/^(0x)[0-9a-f]{5}$/i));
+        });
+
+        it('generates a random hex string with no prefix', function() {
```
vnq, line 244:
```suggestion
        it('generates a random hex string without prefix', function() {

```

maximki: I'm no English major, but I believe that my version is a bit better

PR #1088, opened 2021-01-18
Adds random date generation function

TronSkywalker:
In my codebase, I have multiple instances of this expression. 

`new Date(faker.random.number()) ` to generate a random stream of dates.

The expression is overall useful because óne can strictly require that the faked dates do not change from seed to seed. 

I found it personally to be more practical to have it on the lib side.

### DIFF
--- a/lib/date.js
+++ b/lib/date.js
@@ -164,6 +164,17 @@ var _Date = function (faker) {
       return faker.random.arrayElement(source);
   };
 
+  /**
+   * random
+   *
+   * @method faker.date.random
+   */
+  self.random = function () {
+    var date = new Date(faker.random.number());
+
+    return date
+};
+  
   return self;
 
 };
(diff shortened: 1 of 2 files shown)


TronSkywalker:
Seems like a Bug in the litecoin generator:

`assert.ok(litecoinAddress.match(/^[LM3][a-km-zA-HJ-NP-Z1-9]{25,32}$/));
`

Fixes the test for me, with the same explanation as in the bitcoinAdresses. But I don't know much about the correctness of the coin addresses...

Marak:
@TronSkywalker 

I think we can we do better than `new Date(faker.random.number())` because we already have `date.future`, `date.recent`, `date.between` and `date.past`.

Due to default behavior of `random.number`, doing `new Date(faker.random.number())` always seems to turn a date from the 1970s.

If anything, `date.random` should probably be a composite function containing the other date methods I previously mentioned here. Open to suggestions on how we can improve this.

Don't worry about the litecoin test failure. It's seed related and should be resolved.

PR #1090, opened 2021-01-28
Branch1 : added the new category of musical instruments

neha-89:


Marak:
@neha-89 -

Can you move this new functionality into the `Music` module and make a new method `Music.instrument`?

We don't want to add a new Module just for instruments.

Thanks again!

neha-89:
sure I will add it  in music.. Thank you

PR #1095, opened 2021-02-05
LAPTOPS

VishalGithub11:


Marak:
Thank you for your contribution @VishalGithub11 

Instead of `Laptop`, can we rename this to `Electronics` and add a few more categories of `Electronics`?

 - `Electronics.phone`
 - `Electronics.computer`
 - `Electronics.laptop`
 - `Electronics.homeAssistant`
 - `Electronics.cables`

etc...

VishalGithub11:
Ya sure. That 's an great Idea!!!
Thanks for reply👍(didn't expect  that)

Laptops
```diff
@@ -0,0 +1,127 @@
...
+    ).toUpperCase();
+};
+
+self.vrm.schema = {
+    "description": "Generates a Laptop vrm",
+    "sampleResults": ["MF56UPA", "GL19AAQ", "SF20TTA"]
+};
+};
+
+module["exports"] = Laptop;
```
nurlansu, line 127: File's extension lost. Also file should be placed under `lib/` directory.

nurlansu (changes requested):

Laptops
```diff
@@ -0,0 +1,127 @@
...
+        "sampleResults": ["4GB/1TB", "6GB/1TB", "8GB/1TB", "10GB/1TB"]
+    };
+
+    /**
+     * fuel
+     *
+     * 
+     */
+
+};
```
nurlansu, line 72: You have closed it too soon. `self.vin`, `self.color` and the rest are out of the scope of Laptop.

PR #1103, opened 2021-02-09
Create fashion.js

SanthoshBommideni:


Marak:
@SanthoshBommideni -

We can't merge this in it's current form. There is currently not enough new data in the Pull Request to warrant a new Module. There is also a lot of duplicate methods here we already have elsewhere in `faker`.

Maybe we can try again with a `Clothing` module and add all types of `shirt` and `pants` and `jackets`, etc. All the garment types.

Thanks!

PR #1109, opened 2021-02-13
adding animal module

RavsCodes20:
started with dogs, added a defintion, an array of dogs breeds, and an associated test. Having trouble getting the test to recognize animal as a module. Not sure where else i need to define it.


lib/index.js
```diff
@@ -104,6 +107,7 @@ function Faker (opts) {
   var _definitions = {
     "name": ["first_name", "last_name", "prefix", "suffix", "gender", "title", "male_prefix", "female_prefix", "male_first_name", "female_first_name", "male_middle_name", "female_middle_name", "male_last_name", "female_last_name"],
     "address": ["city_prefix", "city_suffix", "street_suffix", "county", "country", "country_code", "country_code_alpha_3", "state", "state_abbr", "street_prefix", "postcode", "postcode_by_state", "direction", "direction_abbr", "time_zone"],
+    "animal": ["dog", "cat", "snake"],
```
Marak, line 110: Better to not include `cat` and `snake` references if data isn't included yet.

I would go remove these references and we can add `cat` and `snake` in the next PR.

Marak:
Thank you for your contribution @RavsCodes20 

I went ahead and made a few changes to the PR so it would pass the tests and work.

Merged.

PR #1117, opened 2021-02-20
Italian last names extension

supix:
They are 2170 now (were 635 before).


Marak:
Thank you for your contributions @supix 

Looks good to me!

Merged.

Issue #1114, opened 2021-02-17
Module Request: `dataTypes`


Marak:
There should be a new top-level module called `dataTypes`.

It would contain methods such as:

 - `Number`
 - `String`
 - `Boolean`
 - `DateTime`
 - `Float`
 - `UUID`
 - `Hexadecimal`
 - `JSON`
 - `Array`

Most of these methods already exist in `Random` module, so we'll want to move those methods to new `dataTypes`. 

We'll want to put in placeholder functions in `Random` for a major version and put in deprecation `console.log` messages so that users don't get a broken API.

lbuerste:
Hi Marak and all others,
I'm brand new to contributing to opensource, but I'm very eager to finally make a contribution. Because this doesn't seem like an issue too complicated, I'll try to figure it out in the next few days and hopefully opening a pull request! :)

Marak:
Awesome @lbuerste 

Looking forward to it!

Marak:
This should be resolved with #1123 

Thank you for your contributions @lbuerste 

Awesome job!

PR #1123, opened 2021-03-03
Issue 1114 new datatype module

lbuerste:
As described in Issue #1114 , I've implemented a new datatype module. As also mentioned in the issue, many of the "new" methods have been taken from the random module, In summary:

- number, float, uuid, hexaDecimal, boolean have been moved to faker.datatype
- Respective methods in faker.random now call the methods in faker.datatype and log to console "DeprecationWarning: Method is now located in faker.datatype.<method>
- implemented methods for string, datetime, json and array

The current issues from my point of view is, that I was unable to make datetime work with seeding

Furthermore, because the issue didn't describe much about what the new methods should exactly do and which parameters would have been nice to offer, I'd ask you all for some feedback about possible improvements the new methods should get.

Another problem I've had during building was that gulp jsdoc didn't work because it used "README.md" but "Readme.md" was present. Since I didn't change that initially, I don't if thats a problem only I ran into and therefore its not part of the pull request.

### DIFF
--- a/build/src/docs.md
+++ b/build/src/docs.md
@@ -99,12 +99,12 @@ If you want consistent results, you can set your own seed:
 ```js
 faker.seed(123);
 
-var firstRandom = faker.random.number();
+var firstRandom = faker.datatype.number();
 
 // Setting the seed again resets the sequence.
 faker.seed(123);
 
-var secondRandom = faker.random.number();
+var secondRandom = faker.datatype.number();
 
 console.log(firstRandom === secondRandom);
 ```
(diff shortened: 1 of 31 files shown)


Marak:
At first glance this looks really good @lbuerste !!!

I'm going to pull this now and start testing for merge.

Marak:
@lbuerste -

This was merged the other day but for some reason the changes aren't showing here.

It should be in `master`.

Please review if you can and let us know if it's working. I made a few minor changes.

Thanks again for your contributions!

PR #1124, opened 2021-03-03
Hexadecimal cleanup

vnq:
* Added an optional `prefix` parameter to `hexaDecimal` which defaults to true, to keep it backwards compatible.
* Use 1 source for hexadecimal characters for several usages (internet, git, finance).
* Removed the uppercase characters as the tests ignored them anyway and the only existing `hexaDecimal` usage used a `toLowerCase()`
* Made tests more strict

Other thoughts: make upper-, lower- & mixed case options available.


vnq:
Closing in favour of #1083

Issue #1110, opened 2021-02-15
Feature Request: Adding pc parts


LockBlock-dev:
It could be a great idea to add PC parts, like random CPU, GPU, OS.. something like that.

Marak:
Sounds good @LockBlock-dev 

Will accept a Pull Request for this.

LockBlock-dev:
I'm working on it 👌

PR #1125, opened 2021-03-04
added computer module

LockBlock-dev:
PR for [#1110](https://github.com/Marak/faker.js/issues/1110)

I hope it's good, tell me if I need to do some changes.


lib/computer.js
```diff
@@ -0,0 +1,297 @@
...
+    };
+
+
+
+    /**
+    * psu_brand
+    *
+    * @method faker.computer.psu_brand
+    */
+    self.psu_brand = function () {
```
Marak, line 285: All API method names should use camelCase syntax and not contain any `_` characters.

lib/computer.js
```diff
@@ -0,0 +1,297 @@
...
+    };
+
+
+
+    /**
+     * operating_system
+     *
+     * @method faker.computer.operating_system
+     */
+    self.operating_system = function () {
```
Marak, line 269: All API method names should use camelCase syntax and not contain any `_` characters.

lib/computer.js
```diff
@@ -0,0 +1,297 @@
...
+    };
+
+
+
+    /**
+    * motherboard_brand
+    *
+    * @method faker.computer.motherboard_brand
+    */
+    self.motherboard_brand = function () {
```
Marak, line 253: All API method names should use camelCase syntax and not contain any `_` characters.

Marak (changes requested):
@ibroadfo -

This is a really great start!

I've made some code comments about renaming the API method names. Please see other faker modules to see how this should be structured.

In addition, I think we should rename this module to `Electronics`. We have another Pull Request open for `Laptops` right now. We'd like to get all of these merged into `Electronics` module. see: https://github.com/Marak/faker.js/pull/1095

We could add additional methods such as:

```
Electronics.phone
Electronics.computer
Electronics.laptop
Electronics.homeAssistant
Electronics.cables
```

LockBlock-dev:
I will rewrite the methods in camelCase syntax and rewrite the module to replace Computer with Electronics.

PR #1126, opened 2021-03-06
Added italian city names

Cload:
This commits adds a list of italian city names .


Marak:
Thank you for your contributions @Cload 

Please feel free let us know if you find any other parts of the `it` locale which we can improve.

Merged.

Cload:
I think one possible improvement could be tweaking how random city are selected.

This is the function used to pick a random city at the moment:

https://github.com/Marak/faker.js/blob/d5298db0f5e5c91d10a899c9ce0cb8c4bd4cabbc/lib/address.js#L47-L81


I have taken a look at some locales and i noticed that in each locale folder there is a file named 'city.js' which contains some formats that could be used when picking random city names.

https://github.com/Marak/faker.js/blob/master/lib/locales/it/address/city.js


I think it may be useful if the "city" function picked the random format string from the "city.js" of the locale in use instead of using the hard coded formats like it does at the moment.

If a particular locale does not have its 'city.js' file it could always fallback to the current array of hard coded formats.

If you think the idea is feasible i will be glad to submit a pull request.

In the meantime thank you for your amazing library

Marak:
That sounds good @Cload, will accept a Pull Request for that.

Thank you!

Issue #1097, opened 2021-02-05
Feature Request: Expanded `Vehicles`


Marak:

JasonHong1998:
I would love to add this feature as my first contribution to an open-source project. I'll make a pull request with a small demo and we can further discuss the API.

JasonHong1998:
#1127 link to pull request

laulujan:
Hello, is this issue still open? I would love to help to complete the vehicles module

Marak:
@landretk -

Yes, still open.

laulujan:
Great, I'll work on this and send a PR with a proposal, thanks!!

PR #1127, opened 2021-03-07
Issue #1097: add bicycleType in the vehicle module

JasonHong1998:
Hi Marak! For now, I added a "bicycleType" to the vehicle module to get this API started, and if you approve of my work I can start working on the rest. One thing to note is that there is already a "vehicle_type" API that generates types of cars, so if we wanted to add other types of vehicles I think we should modify the "vehicle_type" to "car_type" in order to make room for i.e. "ship_type."


Marak:
Thank you for your contributions @JasonHong1998 

Instead of `bicyleType` can you rename this to `bicycle`?

JasonHong1998:
@Marak done!

Marak:
Thank you for your contribution @JasonHong1998 

Merged

PR #1136, opened 2021-03-11
use city definitions when generating cities

Cload:
This commits allows formats defined in 'city.js' files to be used when generating random city names.

The problem is that many locales have strings inside 'city.js' files that cannot be used as format strings;
Therefore the randomly generated results may contain the actual strings contained inside 'city.js' files.

https://github.com/Marak/faker.js/blob/master/lib/locales/en/address/city.js#L1-L6

What are these strings actually used for?

If they are not actually used anywhere, would it be ok to correct them (using double moustache syntax)  so they can be used to generate city names?
 
If necessary i could look into a function to validate format strings.

Thank you

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -49,7 +49,11 @@ function Address (faker) {
    * method provided by faker wrapped in `{{}}`, e.g. `{{name.firstName}}` in
    * order to build the city name.
    *
-   * If no format string is provided one of the following is randomly used:
+   *
+   * If no format string is provided one it is randomly picked from the active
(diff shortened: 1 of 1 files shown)


Marak:
@Cload -

Not all of the data points in all locales are connected. We should be using these to generate city names when available.

Cload:
When data points are connected, should i correct the incorrect templates ? e.g. the "en" locale has incorrect templates atm

Marak:
We should probably add some tests for this. City name generation should be improved

PR #1139, opened 2021-03-15
Fixed linter issues.

wolfshirts:
Fixed linter issues for my previous pr. Closing it.

### DIFF
--- a/lib/address.js
+++ b/lib/address.js
@@ -39,7 +39,8 @@ function Address (faker) {
   this.zipCodeByState = function (state) {
     var zipRange = faker.definitions.address.postcode_by_state[state];
     if (zipRange) {
-      return faker.datatype.number(zipRange);
+      // Ensure lengh of zip is 5 chars.
+      return `${faker.datatype.number(zipRange)}`.padStart(5, '0');
     }
     return faker.address.zipCode();
   }
@@ -99,7 +100,7 @@ function Address (faker) {
 
   /**
    * Returns a random city name
-   * 
+   *
    * @method faker.address.cityName
    */
   this.cityName = function() {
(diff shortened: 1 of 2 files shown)


lib/address.js
```diff
@@ -39,7 +39,8 @@ function Address (faker) {
   this.zipCodeByState = function (state) {
     var zipRange = faker.definitions.address.postcode_by_state[state];
     if (zipRange) {
-      return faker.datatype.number(zipRange);
+      // Ensure lengh of zip is 5 chars.
+      return `${faker.datatype.number(zipRange)}`.padStart(5, '0');
     }
     return faker.address.zipCode();
   }
```
jbrown123, line 46: Is this function only used for the US, or by countries that have a 5 digit ZIP code?

wolfshirts: My understanding is that it is only when faker.locale is set to en_us, and when you pass a state into zipCodeByState. The definition for the postcode range is in faker.js/lib/locales/en_US/address/postcode_by_state.js and only includes US states.

lib/address.js
```diff
@@ -39,7 +39,8 @@ function Address (faker) {
   this.zipCodeByState = function (state) {
     var zipRange = faker.definitions.address.postcode_by_state[state];
     if (zipRange) {
-      return faker.datatype.number(zipRange);
+      // Ensure lengh of zip is 5 chars.
+      return `${faker.datatype.number(zipRange)}`.padStart(5, '0');
```
Marak, line 43: We aren't using string interpolation yet, so we can't merge this.

`v6.0.0` will have ES6 support.

owens-ben:
So could we please get this change merged? zipCodeByState is completely useless without this change.

wolfshirts:
Pushed an update to this since I was @'d over it. https://github.com/Marak/faker.js/pull/1226

PR #1140, opened 2021-03-16
Add ko address.postcode format

greatSumini:
added fiveDigit format

- Korean postcode: sixDigits with dash format is deprecated, but remain


Marak:
Thank you for your contribution @greatSumini 

Merged.

PR #1143, opened 2021-03-18
Improve faker.random.arrayElements performance

OleksandrYehorov:
`faker.random.arrayElements` has performance issues when dealing with large arrays.

The following implementation significantly improves performance.

Performance tests for an array of 200000 items:
| Count  | Old implementation | New implementation |
| -      | -                  | -                  |
| 20000  |      1872ms        |       10ms         |
| 40000  |      1643ms        |       9ms          |
| 60000  |      1542ms        |       9ms          |
| 80000  |      1429ms        |       10ms         |
| 100000 |      1270ms        |       12ms         |
| 120000 |      1099ms        |       15ms         |
| 140000 |      1013ms        |       25ms         |
| 160000 |      623ms         |       23ms         |
| 180000 |      333ms         |       22ms         |

Performance tests source code is available [here](https://codesandbox.io/s/faker-performance-test-oxsm4?file=/src/index.js)

### DIFF
--- a/lib/random.js
+++ b/lib/random.js
@@ -82,14 +82,20 @@ function Random (faker, seed) {
         count = 0;
       }
 
-      var arrayCopy = array.slice();
-      var countToRemove = arrayCopy.length - count;
-      for (var i = 0; i < countToRemove; i++) {
-        var indexToRemove = faker.datatype.number({ max: arrayCopy.length - 1 });
-        arrayCopy.splice(indexToRemove, 1);
+      var arrayCopy = array.slice(0);
+      var i = array.length;
+      var min = i - count;
+      var temp;
+      var index;
+
+      while (i-- > min) {
+        index = Math.floor((i + 1) * faker.datatype.float({ min: 0, max: 1 }));
+        temp = arrayCopy[index];
+        arrayCopy[index] = arrayCopy[i];
+        arrayCopy[i] = temp;
       }
 
-      return arrayCopy;
+      return arrayCopy.slice(min);
   };
 
   /**
--- a/lib/random.js
+++ b/lib/random.js
@@ -89,7 +89,7 @@ function Random (faker, seed) {
       var index;
 
       while (i-- > min) {
-        index = Math.floor((i + 1) * faker.datatype.float({ min: 0, max: 1 }));
+        index = Math.floor((i + 1) * faker.datatype.float({ min: 0, max: 0.99 }));
         temp = arrayCopy[index];
         arrayCopy[index] = arrayCopy[i];
         arrayCopy[i] = temp;


Marak:
Very nice @OleksandrYehorov 

Could you share with us how you are running these benchmarks? We've been looking to implement benchmark testing for the library.

OleksandrYehorov:
@Marak Here is the source code of the tests https://codesandbox.io/s/oxsm4

However, I don't think that my tests would be applicable for the whole library.

makepanic:
This seems to change the behavior of arrayElements.
I wanted to see how it behaves when plotted using random rgb pixels.
I generate an array with the numbers 0..255 and use random.arrayElements with the length 3 to get each red green and blue value.

https://gist.github.com/makepanic/8b231c933707fe853213cd18275d4920

master:
master

1143:


It seems like the random selection on master might be broken and this PR fixes it? Previously there were only blue hues. Now it's somewhat random.

Note: I might have made mistakes in the script so please verify it yourself :)

OleksandrYehorov:
@makepanic Interesting observation! I just figured that while the implementation from master preserves the order of elements in the returned array, the new implementation does not.

```javascript
const array = [1, 2, 3, 4, 5, 6];
```
master:
```javascript
faker.random.arrayElements([1, 2, 3], 3); // Always returns [1, 2, 3]
```

PR:
```javascript
faker.random.arrayElements([1, 2, 3], 3); // Can return [2, 3, 1] [1, 3, 2] [1, 2, 3] etc.
```

makepanic:
I'm seeing some cases in the generation of this PR returns an undefined value, e.g.:

lib/random.js
```diff
@@ -82,14 +82,20 @@ function Random (faker, seed) {
...
-        var indexToRemove = faker.datatype.number({ max: arrayCopy.length - 1 });
-        arrayCopy.splice(indexToRemove, 1);
+      var arrayCopy = array.slice(0);
+      var i = array.length;
+      var min = i - count;
+      var temp;
+      var index;
+
+      while (i-- > min) {
+        index = Math.floor((i + 1) * faker.datatype.float({ min: 0, max: 1 }));
```
makepanic, line 92: This can generate an index which is bigger than the array.length

OleksandrYehorov: Thanks for finding this out. Fixed it

makepanic: The adjusted index generation seems to have fixed with the latest commit

makepanic (approved):

Marak:
Thank you for this additional review @makepanic 

The images you generated are super cool! I wonder if we can apply this to the library as a more generic method that would work for other `faker` API methods.

Merged.

PR #1146, opened 2021-03-24
Remove tailing quote

spike008t:
Remove tailing quote on Mauricette


Marak:
@spike008t 

Can you provide any data on why this last name should be changed?

spike008t:
@Marak you're welcome.

Because: 
1. All the last name provided on french local doesn't have any tailing quote, looks like a typo error.
2. And I never ever seen a last name with a tailing quote in France.

Marak:
Seems good to me @spike008t 

Merged.

Thanks!

PR #1147, opened 2021-03-24
Give each faker instance its own random number generator

### LINKED ISSUE #768: Seeding bug?
# How to reproduce

## index.js
```js
require('./f1.js');
require('./f2.js');
```

## f1.js
```js
const Faker = require('faker/lib');
const faker = new Faker();

faker.seed(1);

setInterval(() => {
    console.log('reset');
    faker.seed(1);
}, 3000);
```

## f2.js
```js
const Faker = require('faker/lib');
const faker = new Faker();

faker.seed(1);

setInterval(() => {
    console.log(faker.random.number());
}, 1000);
```
## run
```
node index.js
```

# Expected result
Random number generation should not reset.

# Actual result
Random number generation resets, though it should be isolated because i create two different instances.

Problem could be here:
https://github.com/Marak/faker.js/blob/157c0c052ca86eea110e74aa6c4a85f7bdfa9a6b/lib/random.js#L1-L16

`mersenne` library doesn't create instance every time. You should probably create it yourself:
```js
var mer = require('../vendor/mersenne');
var mersenne = new mer.MersenneTwister19937();
```

this pseudo code solves the problem, but its ugly it could be made better:
```js
var mer = require('../vendor/mersenne');
var mersenne = new mer.MersenneTwister19937;

/**
 *
 * @namespace faker.random
 */
function Random (faker, seed) {
  // Use a user provided seed if it exists
  if (seed) {
    var mersenne = new mer.MersenneTwister19937;
    if (Array.isArray(seed) && seed.length) {
      mersenne.init_by_array(seed);
    }
    else {
      mersenne.init_genrand(seed);
    }
  }
```

cowboyd:
> fixes #768 

Currently, every Faker instance shares the same random number generator instance which means that to seed one faker instance is to seed them all. This is also a pre-requisite for #318 (see https://github.com/Marak/faker.js/issues/318#issuecomment-369658181)

This just extracts the public API of the mersenne twister found in `vendor/mersenne.js` and creates a single instance of it per faker.

This implementation is a tad sub-optimal since calling `seed()` on a faker instance will result in two redundant calls to `mersenne.init_genrand()`: one in `random.js` and the other in `datatype.js`. It would probably be more DRY to create the mersenne instance once in faker and then pass it into to the `Random` and `Datatype` instances, however, this takes the strategy of the smallest possible API change.

### DIFF
--- a/lib/mersenne.js
+++ b/lib/mersenne.js
@@ -1,0 +1,31 @@
+var Gen = require('../vendor/mersenne').MersenneTwister19937;
+
+function Mersenne() {
+  var gen = new Gen();
+  gen.init_genrand((new Date).getTime() % 1000000000);
+
+  this.rand = function(max, min) {
+    if (max === undefined)
+    {
+      min = 0;
+      max = 32768;
+    }
+    return Math.floor(gen.genrand_real2() * (max - min) + min);
+  }
+  this.seed = function(S) {
+    if (typeof(S) != 'number')
+    {
+      throw new Error("seed(S) must take numeric argument; is " + typeof(S));
+    }
+    gen.init_genrand(S);
+  }
+  this.seed_array = function(A) {
+    if (typeof(A) != 'object')
+    {
+      throw new Error("seed_array(A) must take array of numbers; is " + typeof(A));
+    }
+    gen.init_by_array(A, A.length);
+  }
+}
+
+module.exports = Mersenne;
--- a/lib/datatype.js
+++ b/lib/datatype.js
@@ -1,5 +1,3 @@
-var mersenne = require('../vendor/mersenne');
-
 /**
  *
  * @namespace faker.datatype
@@ -7,10 +5,10 @@ var mersenne = require('../vendor/mersenne');
 function Datatype (faker, seed) {
   // Use a user provided seed if it is an array or number
   if (Array.isArray(seed) && seed.length) {
-    mersenne.seed_array(seed);
+    faker.mersenne.seed_array(seed);
   }
   else if(!isNaN(seed)) {
-    mersenne.seed(seed);
+    faker.mersenne.seed(seed);
   }
 
   /**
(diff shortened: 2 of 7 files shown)


lib/mersenne.js
```diff
@@ -0,0 +1,31 @@
+var Gen = require('../vendor/mersenne').MersenneTwister19937;
+
+function Mersenne() {
+  var gen = new Gen();
+  gen.init_genrand((new Date).getTime() % 1000000000);
```
Marak, line 5: Can we now remove these lines from ./vendor/mersenne` now? We shouldn't be too worried about backwards compatible for this.

Marak (changes requested):

Marak:
Thank you for your contributions @cowboyd 

Please see code review.

We should be able to get this merged quickly.

lib/mersenne.js
cowboyd: Done!

cowboyd:
@Marak Removed the old api from `vendor/mersenne.js` with minimal impact. Is there a changelog I should update?

Marak:
Looks good to me @cowboyd 

Merged.

Please let us know if you find anything else we can improve.

Thanks again!

cowboyd:
Nice! 🙌🏻 🙌🏻 🙌🏻 

Glad to help out where I can...

cowboyd:
@Marak Oh yeah, one thing I was curious about is what the release cadence is?

Marak:
Hrmm, the browser build seems to be failing now. I think it's related to this.

Investigating.

Screen Shot 2021-03-25 at 5 27 11 PM

Marak:
Should be resolved with a patch to the examples https://github.com/Marak/faker.js/commit/fac3164beb08d2eb175bb980e9821e36be8464af

Doesn't seem like it should be in issue in core

PR #1158, opened 2021-04-07
Avoid ES6 syntax for now

### LINKED ISSUE #1157: Errors with Uglify after upgrading to 5.5
Hi all,

After upgrading from 5.4.0 to 5.5.2 I am getting errors during my build process complaining about [the line in finance.js that uses "let"](https://github.com/Marak/faker.js/blob/master/lib/finance.js#L114). Here is the error I'm getting:

```
ERROR in request.bundle.js from UglifyJs
Unexpected token: name (formattedString) [./node_modules/faker/lib/finance.js:114,0][request.bundle.js:65428,10]
```

I'm not an expert in such matters but my understanding is that ES6 code will not work with Uglify, and "let" is ES6. I assume I could add some transpilation somewhere that would take care of the problem. But I just thought I would bring it up here in case you were not aware. An argument could be made for using "var" until 6.0.0, so that downstream projects don't have to change their build processes until the next major version.

I did a search for "let" in the codebase and it is only used in 2 places. If this is of interest, I'd be happy to submit a PR. If not and you prefer to continue using ES6, feel free to close this.

brockfanning:
Fixes #1157

### DIFF
--- a/build/gulp-tasks/package.js
+++ b/build/gulp-tasks/package.js
@@ -1,12 +1,12 @@
 // generates a new faker package based on a list of locales
 
-let locale = 'en';
+var locale = 'en';
 
 /*
 
   Here you could override the default directory list of all, and replace it with a list of custom locales like:
 
-  let localeList = ['en', 'en_AU', 'de', 'jp', 'ar'];
+  var localeList = ['en', 'en_AU', 'de', 'jp', 'ar'];
 
 */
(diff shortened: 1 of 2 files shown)


Marak:
Thank you for your contributions @brockfanning 

It's much appreciated.

Merged. We should publish a new version before end of day.

Issue #1155, opened 2021-04-02
faker.random.arrayElements returns array with single null item if passing empty array as argument


benjaminjohnking:
When calling `faker.random.arrayElements`, passing an empty array as an argument results in the function returning an array with a single null element.

Example: `faker.random.arrayElements([]);`
Expected result: `[]`
Actual result: `[null]`

Sample use case is setting up a unit test where a dynamically generated array can have 0 or more items in random order. In the event that the size is 0, the unit test fails because the result is one null item instead of 0 items as expected.

If you need any further info or have any questions, please let me know. Thanks in advance for your help!
Ben

PR #1161, opened 2021-04-12
Fix issue #1155

marcioflaviof:
Trying to fix this issue #1155, but it is my first time so forgive-me if i`m doing something wrong


lib/random.js
```diff
@@ -72,14 +72,18 @@ function Random (faker, seed) {
   this.arrayElements = function (array, count) {
     array = array || ["a", "b", "c"];
 
+    if (array.length == 0 && count === undefined) {
```
Braweria, line 75: I suggest using a strict equality sign here  
```diff
- array.length == 0
+ array.length === 0
```

Braweria (changes requested):

Issue #1096, opened 2021-02-05
Module Request: `word`


Marak:
We'd like to see another top module added for `word`.

This could include the following methods to start:

```
noun
pronoun
verb
adjective
adverb
preposition
conjunction
interjection
```

We don't have to include the entire dictionary, but it would be good to have at around 1,000 nouns and 1,000 verbs to start. Will want a good amount of each other type of word as well.

This would be a great Pull Request for someone looking to contribute to the project.

zerefwayne:
@Marak This sounds like something I needed for my project last year. I would be glad to work on adding this one!

Marak:
Closed with #1172

Thanks @ad-walker !

PR #1162, opened 2021-04-13
add Georgian lorem

vmakharashvili:
I added Georgian words in Georgian locale lorem sequence


Marak:
@vmakharashvili - 

Is that how lorem ipsum works?

What is this list of words you've added? I suspect this list of words might be better suited for the `Word` module?

RE: https://github.com/Marak/faker.js/issues/1096

vmakharashvili:
I think you are right. 
I'll make changes with it.

Marak:
Closing due to inactivity.

PR #1163, opened 2021-04-21
Remove deprecated "random" entries from Readme.md

bkir:
Seeing deprecation messages when trying to use faker.random.x -- Do we need any other updates, perhaps here as well?   http://marak.github.io/faker.js/


Marak:
Good catch @bkir 

Merged.

PR #1165, opened 2021-04-26
Update README about running tests

wreiske:
I couldn't find a makefile to run `make test` so I've updated the README.md to include `npm run test` which successfully runs the tests as defined in package.json [here](https://github.com/Marak/faker.js/blob/e073ace19cbf68857a5731dc3302fda0eb36cf24/package.json#L17).


Marak:
Good catch @wreiske !

Merged.

Issue #1133, opened 2021-03-10
Feature Request: Emojis


jSadoski:
Simple as it sounds, it would be great if there a `faker.internet.emoji()` to generate fake emoji data 🙂

I might have some time in the future to PR this, but for now I'm leaving the suggestion here.

Marak:
It would be nice if we could randomly generate emojis.

I'd want to see this as a few lines of code which used a value range for UTF-8 instead of a large mapping file of all emoji references.

Will accept Pull Request for this.

PR #1166, opened 2021-04-26
[WIP] Feature: Random Emoji Support

wreiske:
Closes: https://github.com/Marak/faker.js/issues/1133

This commit adds some basic support for generating emojis from code spaces.

`faker.internet.emoji()`
😍

image

I've coded this in a way to support more code spaces in the future. You can easily add more by adding the UTF-8 code space to the `emojiCodeBlocks`  array. After some testing, I've decided to stick with just basic emoticons for this pull request.

U+1F600 ... U+1F64F: Emoticons https://en.wikipedia.org/wiki/Emoticons_(Unicode_block)

**I'm making this a work in progress because for some reason returning an emoji breaks the unit tests.**
image
If someone can figure out why this is causing the unit tests to fail, I'd greatly appreciate it! 🙏 

The unit test for emojis utilizes a regex I found here: https://thekevinscott.com/emojis-in-javascript/#conclusion


jSadoski:
This is strange... Even when removing your unit test from `internet.unit.js`, it still fails at `faker.fake functional tests` => `datatype.string()` in `all.functional.js` when `faker.locale === "en_US"` (not sure if that has anything to do with it). Also, running `node_modules/.bin/mocha test/all.functional.js` succeeds, so I really don't know what's different in the context of running all the tests.

Kenta1561:
I looked a bit into this and although I cannot figure out why this commit suddenly causes the unit tests to fail, here is what I found out so far: During the functional test of `datatype.string()`, a random String containing two consecutive closing brackets gets generated. It apparently causes `fake.js` to assume that there is a format string to interpolate (like with e.g. `{{name.firstName}}`).

```javascript
// find first matching {{ and }}
var start = str.search('{{');
var end = str.search('}}');

// if no {{ and }} is found, we are done
if (start === -1 && end === -1) {
  return str;
}
```
Maybe replacing `&&` with `||` might solve this problem as two opening brackets **and** two closing brackets are required for correct interpolation?

Marak:
At a quick glance, it sounds like it might be an issue where the previous random seed value is now generating new data which is causing old tests to break. If that is the case we should try to fix the error. We can also pick a new seed value for the failing test.

PR #1172, opened 2021-05-20
[FEATURE] Request #1096, add module 'word'

ad-walker:
Per feature request #1096, I've added a `faker.word` module which includes the requested methods:

| Method      | Entries (locale: en) |
| ----------- | ----------- |
|` adjective`   | 1,347        |
|` adverb`   | 326        |
|` conjunction`   | 62        |
|` interjection`   | 47        |
|` noun`   | 6,801        |
|` preposition`   | 109        |
|` verb`   | 6,057        |

Each method takes an optional length param and attempts return a word of the desired length . If the length cannot be accommodated, the method defaults to returning a random length word.

Since the methods are basically identical, I took a concise approach to the unit tests. Let me know if you'd prefer the more conventional structure.


Marak:
@ad-walker -

I did some local testing and works great!

Merged.

PR #1177, opened 2021-05-28
feat: add uniqueArray helper function

hsw107:
Hi there!

We use faker to generate our test data, and often times we use unique string values as both labels as well as keys in React.

For that we generally use `faker.random.word` or `faker.lorem.word`. The issue here is that calling this multiple times doesn't ensure uniqueness, and there's a chance that the same string gets picked twice. This then causes the tests to fail, as keys need to be unique.

While we could use something like `uuid`, it doesn't emulate our test data as well (and isn't very human readable)

To solve for that I've added a utility that will return a unique array of strings, based on a source and desired length.
The source can be a `faker.definitions` array, custom string array, or a function that generates a string.

If the desired length is greater than the source's length (or possible output, in the case of a function) it will return as many unique entries as it can. It was important to have this function be
* Random
* Unique

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -216,6 +216,35 @@ var Helpers = function (faker) {
   };
 
   /**
+   * takes an array of strings or function that returns a string
+   * and outputs a unique array of strings based on that source
+   * @example uniqueArray(faker.random.word, 50)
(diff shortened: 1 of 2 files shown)


Marak:
Thank you for your contributions @hsw107 

Merged.

PR #1184, opened 2021-06-11
fix for copy paste error in hr.js

burner:
I guess this file was created by copy pasting de.js and the locate in the class ctor was not changed to hr


Marak:
Good catch @burner 

Merged.

burner:
Thank you for merging. Wasn't me, a program I wrote caught it. Work smart not hard eh.

PR #1186, opened 2021-06-15
movie.js

ashish821:


ashish821:
movies.js

lib/vehicle.js
```diff
@@ -1,88 +1,88 @@
 /**
  *
- * @namespace faker.vehicle
+ * @namespace faker.movie
```
razh, line 3: This modifies `vehicle.js` instead of creating a separate `movie.js`/`movies.js`?

Issue #1180, opened 2021-06-03
Vodafone GmbH gives a virus warning for your picture links


DoktorGauss:
![image](https://user-images.githubusercontent.com/8172816/120637667-7a778800-c46f-11eb-9388-e35d552dbbf6.png)

Marak:
No clue. 

The image url you posted is for http://placeimg.com/640/480/business, which is a url faker can generate.

I'll leave this open to see if anyone else has the same problem with placeimg.com

Braweria:
A thought:
The link above is `HTTP` instead of `HTTPS`. The `HTTPS` version links to this image, which is different to the `HTTP` version.  
https://placeimg.com/640/480/business  

I don't get a warning, like above, but Chrome does tell me, that the connection is not secure, since it is only `HTTP`.
![image](https://user-images.githubusercontent.com/62669899/122443388-d548cd80-cf9f-11eb-8990-86104b07599c.png)  
compared to the other version.  
![image](https://user-images.githubusercontent.com/62669899/122443416-dd087200-cf9f-11eb-9555-ccae5120ba24.png)

And according to the `imageUrl` method `HTTPS` is optional instead of a must. Why was this decision made? Why not have it always `HTTPS`?

Braweria:
A repo demonstrating this: https://codesandbox.io/s/faker-imageurl-dl77y?file=/src/App.js

PR #1188, opened 2021-06-17
feat: protocol defaults to a secure location

Braweria:
Now a user has to actively choose to have an insecure protocol URL.  
Omitting the parameter returns a secure protocol.


Braweria:
Changing this feature stems from this issue: https://github.com/Marak/faker.js/issues/1180

PR #1197, opened 2021-07-01
fix typo

jeanmarcprevost:


Marak:
Good catch @jeanmarcprevost !

Contributions like this really help. Thank you!

Merged.

PR #1199, opened 2021-07-05
Add dash between Rolls Royce vehicle manufacturer

Remimstr:
I ran into issues with this when testing against a codebase which has its own list of vehicle manufacturers. It seems pretty clear from a quick google search that 'Rolls Royce" should indeed be "Rolls-Royce". See this page directly from the manufacturer's website

Screen Shot 2021-07-05 at 10 44 42 AM

Furthermore wikipedia indicates the same: https://en.wikipedia.org/wiki/Rolls-Royce_Motor_Cars

### DIFF
--- a/lib/locales/en/vehicle/manufacturer.js
+++ b/lib/locales/en/vehicle/manufacturer.js
@@ -25,7 +25,7 @@ module["exports"] = [
   "Nissan",
   "Polestar",
   "Porsche",
-  "Rolls Royce",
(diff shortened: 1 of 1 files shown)


Remimstr:
Note that I found the same issue mercedez benz, so I included that update as a new commit in this PR

From the manufacturer's website
Screen Shot 2021-07-08 at 10 09 58 AM

Issue #1204, opened 2021-07-12
Update PlaceIMG categories


piranna:
https://github.com/Marak/faker.js/blob/c9764abd20a198e318367952a94587eda1bac645/lib/image.js#L25

According to https://placeimg.com/ and after some testing, it seems now it would only support `animals`, `architecture`, `nature`, `people` and `tech`.

MilosPaunovic:
@piranna You are right.

I've added a PR #1205  to reflect this change.

MilosPaunovic:
I've closed the PR #1205  from above, because this is a much bigger issue than my current fix.

A lot of functions rely on removed categories from PlaceIMG. I could tackle the entire thing if that is okay with you @Marak and if you think it makes sense?

MickL:
I also noticed that images.cats or images.sports dont work. Maybe use a different image provider?

PR #1205, opened 2021-07-15
[FIX][#1204] Updated PlaceIMG catgories to reflect changes 

MilosPaunovic:
[PlaceIMG](https://placeimg.com/) changed their categories, so this change reflects that.

### DIFF
--- a/lib/image.js
+++ b/lib/image.js
@@ -22,7 +22,7 @@ var Image = function (faker) {
    * @method faker.image.image
    */
   self.image = function (width, height, randomize) {
-    var categories = ["abstract", "animals", "business", "cats", "city", "food", "nightlife", "fashion", "people", "nature", "sports", "technics", "transport"];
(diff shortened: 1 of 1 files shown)


piranna:
For cats there's another specific placeholder, probably we can use it.

https://placekitten.com/

MilosPaunovic:
This is a much bigger issue than my current fix, so this PR will be closed.

A lot of functions rely on removed categories from PlaceIMG. I could tackle the entire thing if that is okay with you @Marak and if you think it makes sense?

MilosPaunovic:
This is a much bigger issue than my current fix, so this PR will be closed.

A lot of functions rely on removed categories from PlaceIMG. I could tackle the entire thing if that is okay with you @Marak and if you think it makes sense?

raffobaffo:
Hi! How's the status of this? I'm also kinda of frustrated by setting `"business"` and getting smiling faces 😅

PR #1209, opened 2021-07-21
Update last_name.js

johnfraney:
Add "Franey" as a last name. It's a good name! Anyone would be lucky to have it.


johnfraney:
@Marak Awesome! Thanks for indulging me

Issue #1200, opened 2021-07-06
faker.random.locale returning wrong code for Czech Republic


agustin-rosas:
According to the Country codes in Wikipedia https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes, the correct locale code in two characters for Czech Republic is `cs`, but the method is returning `cz`.

paskahtjandra:
i can fix this issue, can you please assign me @Marak

Marak:
Good catch @agustin-rosas 

Seems to have snuck in with the wrong code here: https://github.com/Marak/faker.js/pull/381 cc @mandulaj 

https://github.com/Marak/faker.js/pull/1227 seems to be the correct solution.

We'll have to hold off on resolving this until the next major version this Fall ( since fixing locale name will break previous API ).

PR #1227, opened 2021-09-18
change cz to cs

paskahtjandra:
Hello @Marak

I just fixed the code for czech republic to resolve issue #1200 .
Please review the changes and merge this PR.

Thank you.


Marak:
Thank you for your contribution @paskahtjandra 

We will include this change in the next `major` release ( since it will break the previous API )

New `major` version should released before Jan 1st 2021.

PR #1231, opened 2021-09-28
Removing a slur

Thomas-Richards-MoHC:
Hello. I was generating data with this library and found the surname "Gaylord" appearing in my output several times.

I understand that it's a real surname, but in the UK, it's also a slur that people use today.

My concern is that someone may get in trouble with their employer or with their customers for having this in their data.

I was going to use this library to generate example data for a customer, and now I'll have to manually edit this out of my output.


UmairJibran (approved):
This PR has removed the slur word `***lord`

Marak:
The Gaylord family's name is derived from the ancient Norman culture that was established in Britain following the Norman Conquest of island in 1066. Their name originated with an early member who was a happy, joyous, and bold person. The name Gaylord derives from the nickname the galliard, which means the bold or the joyous.

PR #1233, opened 2021-10-05
improve postcode range for en_US -> TX

monkpit:
The original range would only ever result in 1 value (73301) when calling `faker.address.zipCodeByState('TX')`

New range taken from here: http://www.structnet.com/instructions/zip_min_max_by_state.html


Marak:
Good catch @monkpit 

Merged!

PR #1234, opened 2021-10-05
added planets data

srushteeM:
Added code to generate random planets names.

### DIFF
--- a/lib/planets.js
+++ b/lib/planets.js
@@ -1,0 +1,23 @@
+/**
+ *
+ * @namespace faker.music
+ */
+ var Planets = function (faker) {
+    var self = this;
+    /**
+       * genre
+       *
+       * @method faker.music.genre
+       */
+    self.name = function () {
+      return faker.random.arrayElement(faker.definitions.planets.name);
+    };
+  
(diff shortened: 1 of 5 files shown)


Marak:
@srushteeM -

`planets` isn't specific enough to warrant it's own module.

If you can create an `astronomy` module that includes `planets`, we can accept it. 

I would also add other categories like `star` and `galaxy`. Doesn't need to be the entire cosmo, just enough to generate a good selection.

lib/planets.js
```diff
@@ -0,0 +1,23 @@
+/**
+ *
+ * @namespace faker.music
```
marvinIsSacul, line 3:
```suggestion
 * @namespace faker.planet
```

lib/planets.js
```diff
@@ -0,0 +1,23 @@
+/**
+ *
+ * @namespace faker.music
+ */
+ var Planets = function (faker) {
+    var self = this;
+    /**
+       * genre
+       *
+       * @method faker.music.genre
```
marvinIsSacul, line 10: check these out too...

PR #1271, opened 2022-01-02
Fix some spacing characters in in Persian locale

peZhmanParsaee:


Amir-Alipour (commented):
i think that's spaces is better

peZhmanParsaee:
> i think that's spaces is better

They were written incorrectly due to Persian grammar.

Amir-Alipour:
> > i think that's spaces is better
> 
> They were written incorrectly due to Persian grammar.

yeah but it's actually more used among the people

peZhmanParsaee:
> > > i think that's spaces is better
> > 
> > 
> > They were written incorrectly due to Persian grammar.
> 
> yeah but it's actually more used among the people

Softwares should fix that spaces if people enter them incorrectly

